PR #744, opened 2026-04-23
Refactor eval monitor extension to support org specific client credentials for score publishing

menakaj:
## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary by CodeRabbit

* **New Features**
  * Per-organization publisher credential provisioning with dynamic secret management and fallback to static credentials
  * Automatic creation, deletion, and secret regeneration for per-org OAuth clients; publishing parameters now include clientId/secret references

* **Configuration**
  * New Thunder admin API settings to enable dynamic provisioning
  * JWT audience validation widened to support wildcard publisher client matching

* **Security**
  * Publisher requests enforce org-binding checks before publishing scores

* **Database**
  * Added migration and model to store org publisher credentials

* **Tests**
  * Tests for publishing parameter resolution with and without per-org credentials


deployments/single-cluster/values-openbao.yaml
```diff
@@ -43,6 +43,9 @@ server:
 
       # AMP Monitor Evaluation publisher credentials
       bao kv put secret/amp-publisher-client-secret value="amp-publisher-client-secret"
+
+      # AMP System Client credentials (for agent-manager to call Thunder admin API)
+      bao kv put secret/amp-system-client-secret value="amp-system-client-secret"
```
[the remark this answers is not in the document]
menakaj: This is only for quickstart setup. Production deployments will not use this script.

deployments/helm-charts/wso2-amp-thunder-extension/values.yaml
```diff
@@ -200,6 +201,18 @@ thunder:
...
+      clientId: "amp-system-client"
+      clientSecret: "amp-system-client-secret"
+      name: "AMP System Client"
+      description: "System client for agent-manager to provision per-org OAuth apps"
+      grantTypes:
+        - "client_credentials"
+      tokenEndpointAuthMethod: "client_secret_basic"
+      pkceRequired: false
+      publicClient: false
+      accessTokenValidityPeriod: 3600
```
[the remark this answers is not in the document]
menakaj: This is for local/ quick start setup only. Production installation guide will have detailed instructions.

agent-manager-service/clients/thundersvc/client.go
```diff
@@ -0,0 +1,390 @@
...
+		baseURL:      baseURL,
+		clientID:     clientID,
+		clientSecret: clientSecret,
+		httpClient:   &http.Client{Timeout: 30 * time.Second},
+	}
+}
+
+// getSystemToken returns a cached system token or fetches a new one.
+func (c *thunderClient) getSystemToken(ctx context.Context) (string, error) {
+	c.mu.Lock()
```
nadheesh: **[Critical] Exclusive mutex held across a full network round-trip.**

`mu.Lock()` is held for the entire duration of `fetchSystemToken`, which is an HTTP call (line 79). Under concurrent load, every goroutine that needs a token serializes here — even goroutines serving completely unrelated orgs. The `singleflight` pattern used in the provisioner above is exactly the right fix here too: take an `RWMutex` read lock for the fast (cached) path, then use a `singleflight` for the slow fetch so only one goroutine does the network call while others wait and share the result without blocking under an exclusive lock.

```go
// Fast path — shared read lock
c.mu.RLock()
if c.cachedToken != "" && time.Now().Before(c.tokenExpiry) {
    t := c.cachedToken
    c.mu.RUnlock()
    return t, nil
}
c.mu.RUnlock()
// Slow path — singleflight
```

agent-manager-service/clients/thundersvc/client.go
```diff
@@ -0,0 +1,390 @@
...
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	respBody, _ := io.ReadAll(resp.Body)
+
+	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusCreated {
+		return "", "", fmt.Errorf("thunder create app returned %d: %s", resp.StatusCode, string(respBody))
+	}
+
+	slog.Info("Thunder created", "status", resp.StatusCode)
```
nadheesh: **[Minor] Log message has no context.**

`"Thunder created"` with only a status code is hard to correlate in production logs. Add at least `"appName"` so it's clear which org's app was created:

```go
slog.Info("Thunder app created", "appName", appName, "status", resp.StatusCode)
```

agent-manager-service/repositories/org_publisher_credential_repository.go
```diff
@@ -0,0 +1,63 @@
...
+	db *gorm.DB
+}
+
+// NewOrgPublisherCredentialRepo creates a new OrgPublisherCredentialRepository
+func NewOrgPublisherCredentialRepo(db *gorm.DB) OrgPublisherCredentialRepository {
+	return &orgPublisherCredentialRepo{db: db}
+}
+
+// GetByOrgName returns the publisher credentials for the given org.
+// Returns (nil, nil) if no record exists; returns (nil, err) on real DB errors.
```
nadheesh: **[Critical] Doc comment contradicts the implementation.**

The comment says *"Returns (nil, nil) if no record exists"* but line 50 returns `(nil, gorm.ErrRecordNotFound)`. Callers in `monitor_executor.go` and `publisher_credential_provisioner.go` correctly use `errors.Is(err, gorm.ErrRecordNotFound)`, so current behaviour is fine — but any future caller following this comment will introduce a silent bug.

Either:
- Fix the comment to say *"Returns (nil, gorm.ErrRecordNotFound) if no record exists"*, or
- Change the implementation to return `nil, nil` on not-found (simpler callers, clearer contract).

agent-manager-service/middleware/jwtassertion/auth.go
```diff
@@ -278,22 +291,42 @@ func validateIssuer(issuer string, allowedIssuers []string) error {
...
-	// Check if any token audience is in the allowed list
+	// Check if any token audience matches an allowed entry
 	for _, aud := range audiences {
-		if _, ok := allowedMap[strings.TrimSpace(aud)]; ok {
+		trimmed := strings.TrimSpace(aud)
+		if _, ok := exactAllowed[trimmed]; ok {
 			return nil
 		}
+		for _, prefix := range prefixAllowed {
+			if strings.HasPrefix(trimmed, prefix) {
```
nadheesh, line 346: **[Critical] Wildcard audience matching broadens the JWT attack surface for all routes, not just the publisher path.**

The allowed audience list in `values.yaml` is now `amp-publisher-*`. Any token with an audience starting with `amp-publisher-` passes here — including tokens for apps created by third parties in Thunder that happen to share the prefix. This check applies to the general `JWTAuthMiddleware`, not only the publisher middleware.

The `isValidPublisherClient` subject check provides a second layer for the publish endpoint, but other routes protected only by `JWTAuthMiddleware` rely solely on this function. Please document this as an explicit security trade-off in the PR description and confirm no other route accepts a `amp-publisher-*` audience without a further subject check.

agent-manager-service/services/publisher_credential_provisioner.go
```diff
@@ -0,0 +1,282 @@
...
+		return ref.Data[0].RemoteRef.Key, ref.Data[0].RemoteRef.Property, nil
+	}
+
+	return "", "", fmt.Errorf("SecretReference %s has no data sources", secretRefName)
+}
+
+// EnsureCredentials provisions per-org publisher credentials.
+// Uses singleflight to serialize the slow provisioning path per orgName.
+func (p *publisherCredentialProvisioner) EnsureCredentials(ctx context.Context, orgName, orgUUID string) (*PublisherCredentials, error) {
+	p.logger.Info("EnsureCredentials called", "orgName", orgName, "orgUUID", orgUUID)
```
nadheesh: **[Minor] Info-level log fires on every call, including the in-memory cache hit path.**

`EnsureCredentials` is called on every monitor creation. At Info level this will flood production logs. The cache hit is the hot path — it should be `Debug`:

```go
p.logger.Debug("EnsureCredentials called", "orgName", orgName)
```
Log at Info only when actually provisioning (inside `provisionCredentials`).

agent-manager-service/services/publisher_credential_provisioner.go
```diff
@@ -0,0 +1,282 @@
...
+		"orgName", orgName, "secretRefName", secretRefName, "dataSources", len(ref.Data))
+
+	for _, ds := range ref.Data {
+		if ds.SecretKey == "client-secret" {
+			return ds.RemoteRef.Key, ds.RemoteRef.Property, nil
+		}
+	}
+
+	if len(ref.Data) > 0 {
+		return ref.Data[0].RemoteRef.Key, ref.Data[0].RemoteRef.Property, nil
```
nadheesh: **[Medium] Unsafe fallback: returns the first data source when `"client-secret"` key is not found.**

If the `SecretReference` has data sources but none is named `"client-secret"`, this silently picks `ref.Data[0]`, which could be a completely unrelated KV path. The workflow will then fail later with an opaque credentials error that is very hard to debug.

This should be a hard error:
```go
return "", "", fmt.Errorf("SecretReference %s has no data source with key \"client-secret\"", secretRefName)
```

agent-manager-service/services/publisher_credential_provisioner.go
```diff
@@ -0,0 +1,282 @@
...
+	p.mu.RLock()
+	if creds, ok := p.cache[orgName]; ok {
+		p.mu.RUnlock()
+		p.logger.Debug("Returning cached publisher credentials", "orgName", orgName)
+		return creds, nil
+	}
+	p.mu.RUnlock()
+
+	// Serialize the DB lookup + provisioning path per orgName via singleflight
+	result, err, _ := p.sfg.Do(orgName, func() (interface{}, error) {
```
nadheesh: **[Medium] `singleflight.Do` does not propagate context cancellation to waiting callers.**

If the first caller's context times out, the in-flight provisioning continues to completion, and all other callers sharing this singleflight key will receive that result (whether success or failure) regardless of their own context deadlines.

Consider `singleflight.DoChan` with a `select` so individual callers can bail:
```go
ch := p.sfg.DoChan(orgName, func() (interface{}, error) {
    return p.provisionCredentials(ctx, orgName, orgUUID)
})
select {
case res := <-ch:
    if res.Err != nil { return nil, res.Err }
    return res.Val.(*PublisherCredentials), nil
case <-ctx.Done():
    return nil, ctx.Err()
}
```

agent-manager-service/services/publisher_credential_provisioner.go
```diff
@@ -0,0 +1,282 @@
...
+	clientID, clientSecret, created, err := p.thunderClient.EnsurePublisherApp(ctx, orgName, orgUUID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to provision Thunder app for org %s: %w", orgName, err)
+	}
+	p.logger.Info("Thunder EnsurePublisherApp result",
+		"orgName", orgName, "clientID", clientID, "created", created, "hasSecret", clientSecret != "")
+
+	// If app already existed in Thunder but not in DB, clientSecret is empty.
+	// Delete and recreate.
+	if !created && clientSecret == "" {
```
nadheesh, line 201: **[Medium] Delete-and-recreate is a destructive silent recovery that can break in-flight jobs.**

If a Thunder app exists but is absent from the local DB (e.g., after a DB restore or partial failure), this path silently deletes and recreates the OAuth app. Any evaluation job currently running that holds a valid token issued by the old app will immediately lose its credentials mid-flight.

This recovery path should:
1. Be gated behind an explicit config flag (`allowRecreateOnSecretLoss`) so operators opt-in.
2. Emit a warning metric / structured log at a higher severity than `Warn` so on-call is alerted.
3. Or — safer — return an error and require manual intervention: the operator can manually delete the stale app and retry.

agent-manager-service/services/monitor_manager.go
```diff
@@ -66,6 +68,8 @@ type monitorManagerService struct {
 	monitorRepo            repositories.MonitorRepository
 	scoreRepo              repositories.ScoreRepository
 	encryptionKey          []byte
+	provisioner            PublisherCredentialProvisioner
+	thunderConfig          config.ThunderConfig
```
nadheesh: **[Medium] `thunderConfig` is redundant: the provisioner already encapsulates Thunder mode.**

The struct stores raw `config.ThunderConfig` solely to check `thunderConfig.BaseURL != ""` at line 172. This creates a maintenance hazard — the mode check is now split between the provisioner (which knows its own mode) and the service (which duplicates the check).

Consider adding a `IsThunderMode() bool` method to `PublisherCredentialProvisioner`, or moving the `orgUUID` validation inside `EnsureCredentials` itself, so the service never needs to inspect the raw Thunder config.

agent-manager-service/controllers/monitor_scores_publisher_controller.go
```diff
@@ -64,6 +69,24 @@ func (c *monitorScoresPublisherController) PublishScores(w http.ResponseWriter,
...
+	// In Thunder mode (multi-tenant), ouHandle is required. In on-prem mode, skip the check.
+	claims := jwtassertion.GetTokenClaims(r.Context())
+	if claims != nil {
+		if c.thunderEnabled {
+			if claims.OuHandle == "" {
+				log.Warn("Missing ouHandle in token while Thunder is enabled", "monitorId", monitorID)
+				utils.WriteErrorResponse(w, http.StatusForbidden, "insufficient permissions")
+				return
+			}
+			if err := c.scoresService.ValidatePublisherOrg(monitorID, claims.OuHandle); err != nil {
```
nadheesh, line 84: **[Medium] `ValidatePublisherOrg` issues a second DB query for a monitor that `PublishScores` will fetch again.**

`ValidatePublisherOrg` fetches the monitor by ID to compare `OrgName` (see `monitor_scores_service.go`). The subsequent `scoresService.PublishScores` call fetches the same monitor a second time. This is two DB round-trips per publish request.

Move the org-binding check into `PublishScores` (or a dedicated service method that receives the already-fetched monitor) to consolidate into a single DB lookup.

deployments/helm-charts/wso2-amp-thunder-extension/values.yaml
```diff
@@ -200,6 +201,21 @@ thunder:
       pkceRequired: false
       publicClient: false
       accessTokenValidityPeriod: 3600
+    # AMP System Client configuration (for agent-manager to provision per-org OAuth apps via Thunder admin API)
+    # WARNING: clientSecret below is a LOCAL-DEVELOPMENT-ONLY placeholder.
+    # In production, override ampSystemClient.clientSecret via a Kubernetes Secret
+    # or external secret manager. Deploying with this default value is insecure.
+    ampSystemClient:
+      clientId: "amp-system-client"
+      clientSecret: "amp-system-client-secret"
```
nadheesh, line 210: **[Critical] Real default credential committed to the chart.**

`clientSecret: "amp-system-client-secret"` is a known-public value. The warning comment above it is not sufficient — operators who deploy with `helm install` and forget to override this value will run with a compromised credential that grants Administrator-level access to Thunder.

Best practice: leave the default empty (`clientSecret: ""`), add a Helm `required` guard or a `NOTES.txt` warning, and fail the bootstrap script if the value is empty. Alternatively move the secret out of `values.yaml` entirely and require `existingSecret`.

deployments/helm-charts/wso2-amp-thunder-extension/templates/amp-thunder-bootstrap.yaml
```diff
@@ -639,3 +639,153 @@ data:
...
+    # Look up the default auth flow ID
+    RESPONSE=$(thunder_api_call GET "/flows?flowType=AUTHENTICATION&limit=200")
+    HTTP_CODE="${RESPONSE: -3}"
+    BODY="${RESPONSE%???}"
+    if [[ "$HTTP_CODE" != "200" ]]; then
+      log_error "Failed to fetch authentication flows (HTTP $HTTP_CODE)"
+      echo "Response: $BODY"
+      exit 1
+    fi
+    AUTH_FLOW_ID=$(echo "$BODY" | sed 's/},{/}\n{/g' | grep '"handle":"default-basic-flow"' | grep -o '"id":"[^"]*"' | head -1 | cut -d'"' -f4)
```
nadheesh, line 678: **[Minor] Fragile JSON parsing with `sed`/`grep` instead of `jq`.**

This bootstrap script parses Thunder API responses using `sed 's/},{/}\n{/g'` which breaks on pretty-printed JSON, deeply nested objects, or any whitespace variation. Every other bootstrap script in this chart that was already present uses `jq`.

Replace with `jq`:
```bash
AUTH_FLOW_ID=$(echo "$BODY" | jq -r '.[] | select(.handle=="default-basic-flow") | .id' | head -1)
```
This is consistent with the rest of the bootstrap scripts and robust against formatting changes.

nadheesh (commented):
Thorough review of the per-org publisher credential provisioning feature. Several items below ranging from critical correctness issues to minor nits. Overall the architecture is solid — the two-mode design and singleflight pattern are well-chosen. The issues below should be addressed before merge.

agent-manager-service/clients/thundersvc/client.go
```diff
@@ -0,0 +1,551 @@
...
+		return "", fmt.Errorf("thunder put app response decode: %w", err)
+	}
+
+	if len(result.InboundAuth) == 0 || result.InboundAuth[0].Config.ClientSecret == "" {
+		return "", fmt.Errorf("thunder put app response missing clientSecret")
+	}
+
+	slog.Info("Thunder client secret regenerated", "appID", appID)
+	return result.InboundAuth[0].Config.ClientSecret, nil
+}
```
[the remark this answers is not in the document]
menakaj: There is no dedicated endpoint for client secret regeneration in thunder.

agent-manager-service/middleware/jwtassertion/auth.go
menakaj: This was fixed with integrating proper org validation with the aud

deployments/helm-charts/wso2-amp-thunder-extension/templates/amp-thunder-bootstrap.yaml
menakaj: Current script is not using jq. to do this need to add jq dependency to scripts and verify properly. Hence let's keep the existing pattern.

deployments/helm-charts/wso2-amp-thunder-extension/values.yaml
menakaj: This is the followed approach for the other secrets in the values.yaml. Fixing this should be separate effort. will create an issue to track this. https://github.com/wso2/agent-manager/issues/750

agent-manager-service/services/publisher_credential_provisioner.go
menakaj: Changed to re-generate the secret if the app is found. The secret will be propagated to existing monitors via secrets since all secrets referencing the entry in secret manager.

agent-manager-service/controllers/monitor_scores_publisher_controller.go
menakaj: PublishScores does not fetch the monitor. It works directly with monitorID and runID — no GetMonitorByID call. The comment in the controller (lines 74-75) already documents this:

  ▎ Note: ValidatePublisherOrg fetches the monitor by ID (single DB query). PublishScores below does NOT re-fetch the monitor — it works directly with monitorID and runID.

agent-manager-service/middleware/jwtassertion/auth.go
```diff
@@ -86,23 +88,54 @@ var (
...
+			// validate that the org handle embedded in the subject matches the ouHandle claim.
+			if claims.Sub != publisherStaticClient {
+				orgHandle := strings.TrimPrefix(claims.Sub, publisherSubjectPrefix)
+				if claims.OuHandle == "" || claims.OuHandle != orgHandle {
+					slog.Warn("Publisher token org mismatch",
+						"sub", claims.Sub, "ouHandle", claims.OuHandle, "expectedOrg", orgHandle)
+					utils.WriteErrorResponse(w, http.StatusForbidden, "insufficient permissions")
+					return
+				}
+			}
```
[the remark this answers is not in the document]
menakaj: 1. The org handle "client" is not a realistic org name — org handles in Thunder/Asgardeo follow organizational naming conventions. An org literally named "client" is contrived.
  2. Defense-in-depth already exists — Even if the middleware skips its subject-vs-OuHandle check, the controller at lines 78-88 enforces org-binding when Thunder is enabled. It requires OuHandle != "" and
  validates it against the monitor's org via ValidatePublisherOrg. So the actual org-binding is still enforced.
  3. The attack requires a malformed token from Thunder itself — For this to be exploitable, Thunder would need to issue a token with sub: "amp-publisher-client" but a wrong OuHandle. If the attacker controls
  Thunder token issuance, you have bigger problems.
  4. Changing the middleware signature to accept a config flag adds complexity — PublisherClientAuthMiddleware() currently takes no args and is cleanly composed. Adding thunderEnabled bool just for this edge case
  couples middleware to deployment config for a scenario that's already covered by the controller.

  The middleware's role here is subject format validation and basic consistency checking. The controller is where the real tenant-isolation enforcement happens, and it already does that correctly.

nadheesh (approved):

PR #789, opened 2026-05-02
Add Api-Key Management functionality

AnoshanJ:
## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Release Notes

* **New Features**
  * Added API key management for agents: create, revoke, and list API keys
  * Added endpoint authentication configuration for agent deployments
  * Introduced Agent Security page to manage API keys with UI controls
  * Added test API key generation for development and testing
  * Added API key support to agent test interfaces (Chat and Swagger)

* **Infrastructure Updates**
  * Renamed gateway from AI Gateway to API Platform Gateway
  * Updated gateway default ports and configurations
  * Enhanced gateway management with authentication credential support


menakaj:
**Fundamental question.**

This implementation assumes that the gateways which are registered will always be accessible from the agent manager service. And we manage each gateway's api access. 

However, as we designed the gateway management features earlier, it was decided that, gateways can run anywhere (whether agent manager may or may not have direct network access), and gateways will be connected to the agent manager service via websocket connection. All the apis, api keys will be propagated to the gateway via the web socket connection. Restarting, re-initiating the gw will first connect to the service and fetch the apis, and api keys.

With this new implementation however, change that model. 

There are 2 approaches that we can take here.

1. Use OC traits with pre-defined api names
In OC api management trait, we can provide the api name. In this case, we can always generate the api name for a given agent. Then use the same name to create the api key against. (if api keys are created for api name/ api handle)

2. Skip using Traits and manage the api separately.
In this approach, we do not use oc traits, instead we  create the rest api object for the agent and deploy to the env's gateway via websocket channel. And here we manage every parameter of the api. 

I think we have to pick one of the above and do the implementation instead of invoking each gateway's api.

PR #790, opened 2026-05-02
Support per-org publisher credentials for multi-tenant monitor scheduling

### LINKED ISSUE #782: Fix monitor operations when using Thunder IDP: use org-scoped tokens for OpenChoreo API calls
## Problem

When the AMP platform is deployed with Thunder as the identity provider (multi-tenant mode), all monitor operations fail because the OpenChoreo API requires tokens with an organization subscription claim (`org sub`). The current implementation uses a single system-level token (from `cfg.IDP.ClientID/ClientSecret`) for all OpenChoreo API calls, which does not carry the org sub and is therefore rejected.

## Root Cause

Per-org publisher credentials (client_id + client_secret) are provisioned via Thunder and stored partially in the `org_publisher_credentials` DB table. The `client_secret` is stored only in OpenBao (referenced by `secret_kv_path`/`secret_key`), which the agent-manager-service cannot read back at runtime. Without the secret, the service cannot generate org-scoped tokens.

## Affected Operations

- Monitor creation (`CreateMonitor`) — resolves agent/environment via OC API, creates LLM proxy SecretReference
- Monitor update (`UpdateMonitor`) — re-provisions LLM proxy via OC API
- Past monitor runs — triggered during creation via `ExecuteMonitorRun`
- Monitor execution (`ExecuteMonitorRun`) — creates WorkflowRun CR in OpenChoreo
- Future monitor scheduler (`triggerMonitor`) — calls ExecuteMonitorRun on schedule

## Proposed Fix

1. Store the `client_secret` encrypted in the `org_publisher_credentials` table using AES-256-GCM (same pattern used for LLM provider credentials).
2. Add `GetOCClientForOrg(ctx, orgName)` to the `PublisherCredentialProvisioner` — decrypts the stored secret, generates an org-scoped token via the IDP token endpoint (reusing `ocauth.NewAuthProvider`), and returns an OC client configured with that token. Auth providers are cached per-org to avoid repeated token fetches.
3. In monitor create/update/execute operations, use the org-scoped OC client instead of the system OC client when Thunder mode is active.

## Files to Change

- `db_migrations/014_add_encrypted_secret_to_publisher_creds.go` (new)
- `db_migrations/migration_list.go`
- `models/monitor.go`
- `repositories/org_publisher_credential_repository.go`
- `services/publisher_credential_provisioner.go`
- `services/monitor_manager.go`
- `services/monitor_executor.go`
- `wiring/wire.go`

nadheesh:
## Summary

Closes #782

In Thunder (multi-tenant) mode, tokens are org-scoped — a single shared token cannot list or manage monitors across all orgs. This PR makes the monitor scheduler work correctly in that environment by provisioning a per-org OAuth app via Thunder and using org-bound tokens for all OpenChoreo API calls.

- **Per-org credential provisioning** — `PublisherCredentialProvisioner` provisions a Thunder OAuth app for each org on monitor creation, stores the encrypted client secret in the DB, and exposes `GetOCClientForOrg` so the scheduler can obtain org-scoped tokens at runtime without a live user JWT
- **DB migration** — adds `client_secret_encrypted` column to `org_publisher_credentials` (migration 014)
- **Scheduler multi-tenant support** — `orgOCClient()` returns the org-bound OC client in Thunder mode; falls back to the system client otherwise
- **ClusterAuthzRole Helm template** — defines `amp-monitor-scheduler` with `workflowrun:create` + `workflowrun:view`; each org's publisher app is bound to this role via `EnsureClusterRoleBinding`
- **Frontend** — replaces the credential form with an org-level LLM provider picker; updates the monitor API contract (`llmProvider` as a ref instead of an embedded config)

## Bug fixes included

- **Double-trigger prevention** — `UpdateNextRunTime` failure now returns an error so a DB hiccup doesn't leave `next_run_time` stale and cause the next scheduler cycle to fire a duplicate WorkflowRun
- **`tx.Commit()` error now logged** — the advisory lock transaction commit failure was silently swallowed
- **Nil guard in `syncSingleRunStatus`** — defensive check after `GetMonitorByID` to avoid a panic on a deleted monitor
- **Double-checked locking fixed** — `getOrCreateOrgAuthProvider` now uses `singleflight` to prevent redundant auth provider construction under concurrent load
- **Evaluator identifier matching** — `SelectPresetMonitors` now uses `getEvaluatorIdentifier()` on both sides of the config lookup so evaluators resolved via slug fallback still find their saved config

## Test plan

- [ ] Run `make dev-test` in `agent-manager-service/` — all tests pass
- [ ] Create a monitor in Thunder mode and verify the scheduler triggers it using an org-scoped token (not the system token)
- [ ] Stop the DB briefly during a scheduler cycle; confirm no duplicate WorkflowRun is created on recovery
- [ ] Delete a monitor and verify `ExpireWorkflowRun` still succeeds via the system client
- [ ] Verify the `ClusterAuthzRole` CR is created by the Helm chart and that `EnsureClusterRoleBinding` binds the publisher app correctly

## Summary by CodeRabbit

* **New Features**
  * Past monitor evaluations accept traceStart/traceEnd windows (API, UI, and SDK/types).
  * Scheduler supports per-organization provider credentials and org-scoped execution.

* **Improvements**
  * UI datepickers and validation updated (traceStart max 30 days ago; traceEnd must not be in the future).
  * Publisher credentials are now stored encrypted and migration added for the DB.
  * Provisioning enforces required cluster role binding for scheduler; new cluster role grants workflow run create/view.

* **Bug Fixes**
  * Scheduler reliability, error handling, and run-status sync robustness improved.

### DIFF
--- a/agent-manager-service/clients/openchoreosvc/client/authz.go
+++ b/agent-manager-service/clients/openchoreosvc/client/authz.go
@@ -62,7 +62,37 @@ func (c *openChoreoClient) EnsureClusterRoleBinding(ctx context.Context, clientI
 	case http.StatusCreated, http.StatusOK:
 		return nil
 	case http.StatusConflict:
-		// Binding already exists — idempotent success
+		// Binding name already exists — verify it maps the expected clientID → roleName.
+		getResp, err := c.ocClient.GetClusterRoleBindingWithResponse(ctx, bindingName)
+		if err != nil {
+			return fmt.Errorf("failed to get existing ClusterAuthzRoleBinding %s: %w", bindingName, err)
+		}
+		if getResp.StatusCode() == http.StatusOK && getResp.JSON200 != nil {
+			existing := getResp.JSON200
+			if existing.Spec != nil &&
+				existing.Spec.Effect != nil &&
+				*existing.Spec.Effect == effect &&
+				existing.Spec.Entitlement.Claim == "sub" &&
+				existing.Spec.Entitlement.Value == clientID &&
+				len(existing.Spec.RoleMappings) == 1 &&
+				existing.Spec.RoleMappings[0].RoleRef.Kind == roleRefKind &&
+				existing.Spec.RoleMappings[0].RoleRef.Name == roleName {
+				return nil // already correct
+			}
+		}
+		// Spec differs — overwrite with the desired binding.
+		updateResp, err := c.ocClient.UpdateClusterRoleBindingWithResponse(ctx, bindingName, body)
+		if err != nil {
+			return fmt.Errorf("failed to update ClusterAuthzRoleBinding for %s: %w", clientID, err)
+		}
+		if updateResp.StatusCode() != http.StatusOK {
+			return handleErrorResponse(updateResp.StatusCode(), ErrorResponses{
+				JSON400: updateResp.JSON400,
+				JSON401: updateResp.JSON401,
+				JSON403: updateResp.JSON403,
+				JSON500: updateResp.JSON500,
+			})
+		}
 		return nil
 	default:
 		return handleErrorResponse(resp.StatusCode(), ErrorResponses{
--- a/agent-manager-service/clients/openchoreosvc/client/authz.go
+++ b/agent-manager-service/clients/openchoreosvc/client/authz.go
@@ -1,0 +1,75 @@
+// Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+//
+// WSO2 LLC. licenses this file to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package client
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/wso2/agent-manager/agent-manager-service/clients/openchoreosvc/gen"
+)
+
+// EnsureClusterRoleBinding creates a ClusterAuthzRoleBinding for clientID → roleName.
+// Idempotent: a 409 Conflict means the binding already exists and is treated as success.
+func (c *openChoreoClient) EnsureClusterRoleBinding(ctx context.Context, clientID, roleName string) error {
+	effect := gen.ClusterAuthzRoleBindingSpecEffectAllow
+	roleRefKind := gen.ClusterAuthzRoleMappingRoleRefKindClusterAuthzRole
+	bindingName := "amp-publisher-" + clientID + "-scheduler"
+
+	body := gen.ClusterAuthzRoleBinding{
+		Metadata: gen.ObjectMeta{Name: bindingName},
+		Spec: &gen.ClusterAuthzRoleBindingSpec{
+			Effect: &effect,
+			Entitlement: gen.AuthzEntitlementClaim{
+				Claim: "sub",
+				Value: clientID,
+			},
+			RoleMappings: []gen.ClusterAuthzRoleMapping{
+				{
+					RoleRef: struct {
+						Kind gen.ClusterAuthzRoleMappingRoleRefKind `json:"kind"`
+						Name string                                 `json:"name"`
+					}{
+						Kind: roleRefKind,
+						Name: roleName,
+					},
+				},
+			},
+		},
+	}
+
+	resp, err := c.ocClient.CreateClusterRoleBindingWithResponse(ctx, body)
+	if err != nil {
+		return fmt.Errorf("failed to create ClusterAuthzRoleBinding for %s: %w", clientID, err)
+	}
+
+	switch resp.StatusCode() {
+	case http.StatusCreated, http.StatusOK:
+		return nil
+	case http.StatusConflict:
+		// Binding already exists — idempotent success
+		return nil
+	default:
+		return handleErrorResponse(resp.StatusCode(), ErrorResponses{
+			JSON400: resp.JSON400,
+			JSON401: resp.JSON401,
+			JSON403: resp.JSON403,
+			JSON500: resp.JSON500,
+		})
+	}
+}
--- a/agent-manager-service/services/monitor_manager.go
+++ b/agent-manager-service/services/monitor_manager.go
@@ -435,6 +435,21 @@ func (s *monitorManagerService) UpdateMonitor(ctx context.Context, orgName, proj
 			return nil, fmt.Errorf("samplingRate must be between 0 (exclusive) and 1 (inclusive): %w", utils.ErrInvalidInput)
 		}
 	}
+	if monitor.Type == models.MonitorTypePast && (req.TraceStart != nil || req.TraceEnd != nil) {
+		effectiveStart := monitor.TraceStart
+		if req.TraceStart != nil {
+			effectiveStart = req.TraceStart
+		}
+		effectiveEnd := monitor.TraceEnd
+		if req.TraceEnd != nil {
+			effectiveEnd = req.TraceEnd
+		}
+		if effectiveStart != nil && effectiveEnd != nil {
+			if err := validateTraceWindow(effectiveStart, effectiveEnd); err != nil {
+				return nil, err
+			}
+		}
+	}
 
 	// Apply partial updates
 	if req.DisplayName != nil {
@@ -960,20 +975,29 @@ func (s *monitorManagerService) deriveMonitorStatus(monitorType string, nextRunT
 	}
 }
 
+// validateTraceWindow checks the three invariants that apply to both create and update:
+// traceEnd > traceStart, traceEnd not in the future, traceStart within 30 days.
+func validateTraceWindow(traceStart, traceEnd *time.Time) error {
+	if !traceEnd.After(*traceStart) {
+		return fmt.Errorf("traceEnd must be after traceStart: %w", utils.ErrInvalidInput)
+	}
+	if traceEnd.After(time.Now()) {
+		return fmt.Errorf("traceEnd must not be in the future: %w", utils.ErrInvalidInput)
+	}
+	if time.Since(*traceStart) > 30*24*time.Hour {
+		return fmt.Errorf("traceStart cannot be more than 30 days ago: %w", utils.ErrInvalidInput)
+	}
+	return nil
+}
+
 // validateCreateRequest validates the create monitor request based on type
 func (s *monitorManagerService) validateCreateRequest(req *models.CreateMonitorRequest) error {
 	if req.Type == models.MonitorTypePast {
 		if req.TraceStart == nil || req.TraceEnd == nil {
 			return fmt.Errorf("traceStart and traceEnd are required for past monitors: %w", utils.ErrInvalidInput)
 		}
-		if !req.TraceEnd.After(*req.TraceStart) {
-			return fmt.Errorf("traceEnd must be after traceStart: %w", utils.ErrInvalidInput)
-		}
-		if req.TraceEnd.After(time.Now().Add(30 * time.Second)) {
-			return fmt.Errorf("traceEnd must not be in the future: %w", utils.ErrInvalidInput)
-		}
-		if time.Since(*req.TraceStart) > 30*24*time.Hour {
-			return fmt.Errorf("traceStart cannot be more than 30 days ago: %w", utils.ErrInvalidInput)
+		if err := validateTraceWindow(req.TraceStart, req.TraceEnd); err != nil {
+			return err
 		}
 	}
 	if req.IntervalMinutes != nil {
--- a/agent-manager-service/services/monitor_manager.go
+++ b/agent-manager-service/services/monitor_manager.go
@@ -969,8 +969,11 @@ func (s *monitorManagerService) validateCreateRequest(req *models.CreateMonitorR
 		if !req.TraceEnd.After(*req.TraceStart) {
 			return fmt.Errorf("traceEnd must be after traceStart: %w", utils.ErrInvalidInput)
 		}
-		if req.TraceEnd.After(time.Now()) {
+		if req.TraceEnd.After(time.Now().Add(30 * time.Second)) {
 			return fmt.Errorf("traceEnd must not be in the future: %w", utils.ErrInvalidInput)
+		}
+		if time.Since(*req.TraceStart) > 30*24*time.Hour {
+			return fmt.Errorf("traceStart cannot be more than 30 days ago: %w", utils.ErrInvalidInput)
 		}
 	}
 	if req.IntervalMinutes != nil {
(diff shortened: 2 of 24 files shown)


agent-manager-service/clients/openchoreosvc/client/authz.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create ClusterAuthzRoleBinding for %s: %w", clientID, err)
+	}
+
+	switch resp.StatusCode() {
+	case http.StatusCreated, http.StatusOK:
+		return nil
+	case http.StatusConflict:
+		// Binding already exists — idempotent success
+		return nil
```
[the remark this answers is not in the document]
nadheesh: Fixed. On 409, we now GET the existing binding and verify the spec matches the expected `clientID → roleName` mapping. If it matches, we return nil. If it differs (stale binding), we UPDATE it with the desired spec.

agent-manager-service/services/monitor_manager.go
```diff
@@ -969,9 +969,12 @@ func (s *monitorManagerService) validateCreateRequest(req *models.CreateMonitorR
 		if !req.TraceEnd.After(*req.TraceStart) {
 			return fmt.Errorf("traceEnd must be after traceStart: %w", utils.ErrInvalidInput)
 		}
-		if req.TraceEnd.After(time.Now()) {
+		if req.TraceEnd.After(time.Now().Add(30 * time.Second)) {
 			return fmt.Errorf("traceEnd must not be in the future: %w", utils.ErrInvalidInput)
 		}
```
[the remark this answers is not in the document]
nadheesh: Removed the 30s grace entirely — `traceEnd` must now be `≤ time.Now()` for past monitors. The grace window was unintentional.

agent-manager-service/services/publisher_credential_provisioner.go
```diff
@@ -64,22 +75,45 @@ func (s *staticPublisherCredentialProvisioner) EnsureCredentials(_ context.Conte
 
 func (s *staticPublisherCredentialProvisioner) IsThunderMode() bool { return false }
 
+func (s *staticPublisherCredentialProvisioner) GetOCClientForOrg(_ context.Context, orgName string) (client.OpenChoreoClient, error) {
+	return nil, fmt.Errorf("GetOCClientForOrg called in non-Thunder mode for org %s", orgName)
+}
```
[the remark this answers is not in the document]
nadheesh: Fixed. `GetOCClientForOrg` on the static provisioner now returns `nil, nil` per the interface contract.

agent-manager-service/services/publisher_credential_provisioner.go
```diff
@@ -255,3 +310,65 @@ func (p *publisherCredentialProvisioner) provisionCredentials(ctx context.Contex
...
+			ClientSecret: string(secretBytes),
+		})
+		p.orgAuthProviders[orgName] = ap
+		p.logger.Debug("Created org auth provider", "orgName", orgName, "clientID", cred.ClientID)
+		return ap, nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return result.(client.AuthProvider), nil
```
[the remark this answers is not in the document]
nadheesh: Fixed. Using `:` as a separator — `"provision:"` prefix for `EnsureCredentials` and `"auth:"` prefix for `getOrCreateOrgAuthProvider`. Since org names cannot contain `:`, the namespaces are guaranteed disjoint.

console/workspaces/pages/eval/src/subComponents/MonitorFormWizard.tsx
```diff
@@ -211,8 +211,18 @@ export function MonitorFormWizard({
...
+    "traceStart",
+    "traceEnd",
+    "intervalMinutes",
+  ];
+  const page1HasErrors = page1Fields.some((f) => Boolean(errors[f]));
   const canAdvance =
-    formData.displayName.trim().length >= 3 && formData.name.trim().length >= 3;
+    formData.displayName.trim().length >= 3 &&
+    formData.name.trim().length >= 3 &&
+    !page1HasErrors;
```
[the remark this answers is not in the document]
nadheesh: Fixed. `page1Fields` is now built conditionally based on `formData.type` — `"past"` monitors include `traceStart`/`traceEnd`, `"future"` monitors include `intervalMinutes`. Stale errors from hidden fields can no longer block the "Next" button.

agent-manager-service/clients/openchoreosvc/client/authz.go
```diff
@@ -0,0 +1,101 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to get existing ClusterAuthzRoleBinding %s: %w", bindingName, err)
+		}
+		if getResp.StatusCode() == http.StatusOK && getResp.JSON200 != nil {
+			existing := getResp.JSON200
+			if existing.Spec != nil &&
+				existing.Spec.Entitlement.Value == clientID &&
+				len(existing.Spec.RoleMappings) > 0 &&
+				existing.Spec.RoleMappings[0].RoleRef.Name == roleName {
+				return nil // already correct
```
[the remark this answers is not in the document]
nadheesh: Fixed. The spec match now checks all fields for full equality: `Effect`, `Entitlement.Claim`, `Entitlement.Value`, `len(RoleMappings) == 1`, `RoleMappings[0].RoleRef.Kind`, and `RoleMappings[0].RoleRef.Name`. Any deviation triggers an UPDATE.

agent-manager-service/utils/makeresults.go
```diff
@@ -634,6 +634,8 @@ func ConvertToUpdateMonitorRequest(req *spec.UpdateMonitorRequest) *models.Updat
 		LLMProvider:      llmProvider,
 		ClearLLMProvider: clearLLMProvider,
 		IntervalMinutes:  intervalMinutes,
+		TraceStart:       req.TraceStart,
+		TraceEnd:         req.TraceEnd,
```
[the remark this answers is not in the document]
nadheesh: Fixed. Extracted a shared `validateTraceWindow` helper (traceEnd > traceStart, traceEnd not in future, traceStart within 30 days) and called it from both `validateCreateRequest` and `UpdateMonitor`. The update path computes effective start/end by merging the incoming partial values with the existing monitor fields before validating.

console/workspaces/pages/eval/src/EditMonitor.Component.tsx
```diff
@@ -129,6 +129,8 @@ export const EditMonitorComponent: React.FC = () => {
           ? { llmProvider: values.llmProvider ?? null }
           : {}),
         intervalMinutes: values.intervalMinutes ?? undefined,
+        traceStart: values.traceStart?.toISOString(),
+        traceEnd: values.traceEnd?.toISOString(),
```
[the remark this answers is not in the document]
nadheesh: Fixed. In edit mode (`isTypeEditable=false`), `traceStart`/`traceEnd` are excluded from the page-1 navigation gate so stale or expired trace windows don't block advancing to page 2. Validation errors still render inline on the fields, and `guardSubmit` blocks final submission until they are resolved. The date picker's `minDateTime` is kept consistent with the 30-day schema constraint in both create and edit modes.

agent-manager-service/services/publisher_credential_provisioner.go
```diff
@@ -255,3 +313,65 @@ func (p *publisherCredentialProvisioner) provisionCredentials(ctx context.Contex
...
+	return client.NewOpenChoreoClient(&client.Config{
+		BaseURL:      p.ocBaseURL,
+		AuthProvider: authProv,
+	})
+}
+
+// getOrCreateOrgAuthProvider returns (or lazily creates) a cached auth provider for the org.
+// singleflight deduplicates concurrent calls for the same org; the write lock prevents a
+// TOCTOU race between the cache miss check and the map write.
+func (p *publisherCredentialProvisioner) getOrCreateOrgAuthProvider(ctx context.Context, orgName string) (client.AuthProvider, error) {
```
menakaj: #### Write lock held across DB I/O in getOrCreateOrgAuthProvider

```go
result, err, _ := p.sfg.Do("auth:"+orgName, func() (any, error) {
      p.orgAuthMu.Lock()          // write lock acquired here
      defer p.orgAuthMu.Unlock()
      if ap, ok := p.orgAuthProviders[orgName]; ok {
          return ap, nil
      }
      cred, err := p.credRepo.GetByOrgName(orgName)  // DB call under write lock!
      ...
      secretBytes, err := utils.DecryptBytes(...)     // CPU under write lock
      ...
      p.orgAuthProviders[orgName] = ap
      return ap, nil
  })
```

  A write lock is held for the duration of the singleflight function body, which includes a
  database call.
The singleflight already serializes concurrent callers for the same orgName key — the write
  lock is only needed for the map write at the end. 

Fix:

```go
  result, err, _ := p.sfg.Do("auth:"+orgName, func() (any, error) {
      // Re-check under read lock (another goroutine may have populated while we queued)
      p.orgAuthMu.RLock()
      if ap, ok := p.orgAuthProviders[orgName]; ok {
          p.orgAuthMu.RUnlock()
          return ap, nil
      }
      p.orgAuthMu.RUnlock()

      // DB + decrypt with no lock held
      cred, err := p.credRepo.GetByOrgName(orgName)
      ...
      secretBytes, err := utils.DecryptBytes(...)
      ...
      ap := ocauth.NewAuthProvider(...)

      // Write lock only for the map insert
      p.orgAuthMu.Lock()
      p.orgAuthProviders[orgName] = ap
      p.orgAuthMu.Unlock()
      return ap, nil
  })
```

agent-manager-service/services/publisher_credential_provisioner.go
```diff
@@ -255,3 +313,65 @@ func (p *publisherCredentialProvisioner) provisionCredentials(ctx context.Contex
 		SecretKey:    resolvedKey,
 	}, nil
 }
+
+// GetOCClientForOrg returns an OC client authenticated with the publisher app's org-scoped token.
+// Used by the scheduler for CreateWorkflowRun and GetWorkflowRun — operations that run
+// without a live user request context.
+func (p *publisherCredentialProvisioner) GetOCClientForOrg(ctx context.Context, orgName string) (client.OpenChoreoClient, error) {
```
menakaj: #### GetOCClientForOrg creates a new OpenChoreoClient on every call

```go
  func (p *publisherCredentialProvisioner) GetOCClientForOrg(...) (client.OpenChoreoClient,
  error) {
      authProv, err := p.getOrCreateOrgAuthProvider(ctx, orgName)
      ...
      return client.NewOpenChoreoClient(&client.Config{   // new HTTP client every call
          BaseURL:      p.ocBaseURL,
          AuthProvider: authProv,
      })
  }
```

  The AuthProvider is cached per org, but NewOpenChoreoClient is called on every scheduler
  cycle for every org. This creates a new http.Client with a fresh connection pool each time,
  bypassing keep-alive and connection reuse. Cache the OpenChoreoClient in a second map
  alongside orgAuthProviders, using the same lock pattern.

agent-manager-service/services/monitor_scheduler.go
```diff
@@ -298,3 +294,18 @@ func (s *monitorSchedulerService) syncSingleRunStatus(ctx context.Context, run *
 
 	return nil
 }
+
+// orgOCClient returns a per-org OC client in Thunder mode, or the system client in non-Thunder mode.
+func (s *monitorSchedulerService) orgOCClient(ctx context.Context, orgName string) (client.OpenChoreoClient, error) {
```
menakaj, line 298: ####  Dead ErrNotThunderMode branch in orgOCClient


```go  
func (s *monitorSchedulerService) orgOCClient(...) (client.OpenChoreoClient, error) {
      if !s.provisioner.IsThunderMode() {
          return s.ocClient, nil   // returns here for non-Thunder
      }
      orgClient, err := s.provisioner.GetOCClientForOrg(ctx, orgName)
      if errors.Is(err, ErrNotThunderMode) {   // unreachable — already handled above
          return s.ocClient, nil
      }
      ...
  }
```

  ErrNotThunderMode is only returned by
  staticPublisherCredentialProvisioner.GetOCClientForOrg, but IsThunderMode() on the static
  provisioner always returns false, so we return before ever reaching GetOCClientForOrg.
  Remove the dead branch.

agent-manager-service/services/publisher_credential_provisioner.go
```diff
@@ -255,3 +313,65 @@ func (p *publisherCredentialProvisioner) provisionCredentials(ctx context.Contex
...
+		// while we were waiting on singleflight.
+		p.orgAuthMu.Lock()
+		defer p.orgAuthMu.Unlock()
+		if ap, ok := p.orgAuthProviders[orgName]; ok {
+			return ap, nil
+		}
+
+		cred, err := p.credRepo.GetByOrgName(orgName)
+		if err != nil {
+			return nil, fmt.Errorf("no publisher credentials for org %s: %w", orgName, err)
```
menakaj: #### GetByOrgName error not distinguished from "not found"

"Distinguish 'not found' from real errors." A `gorm.ErrRecordNotFound` (org hasn't had EnsureCredentials called yet) produces the same error message as a DB connection failure. The caller gets no signal to distinguish "credentials don't exist, call EnsureCredentials" from "DB is down, retry later." Check for errors.Is(err, gorm.ErrRecordNotFound) and return a distinct typed error.

agent-manager-service/services/publisher_credential_provisioner.go
nadheesh: Fixed in 7608aa8c — the full `OpenChoreoClient` (not just the `AuthProvider`) is now cached per org in `orgOCClients`. The client is built once and reused across scheduler cycles, so the `http.Client` connection pool and token-refresh state are preserved.

agent-manager-service/services/publisher_credential_provisioner.go
nadheesh: Fixed in 7608aa8c — the write lock is no longer held across DB I/O and decryption. The singleflight callback now does a read-lock re-check at entry, then drops all locks for `GetByOrgName` and `DecryptBytes`, and acquires the write lock only for the final map insert.

agent-manager-service/services/monitor_scheduler.go
nadheesh: Fixed in 7608aa8c — the dead `errors.Is(err, ErrNotThunderMode)` branch is removed. `orgOCClient` now returns early when `IsThunderMode()` is false (using the system client), and delegates directly to `GetOCClientForOrg` otherwise.

agent-manager-service/services/publisher_credential_provisioner.go
nadheesh: Fixed in 7608aa8c — `gorm.ErrRecordNotFound` is now explicitly checked and mapped to the new typed `ErrPublisherCredentialNotFound` error. Callers can use `errors.Is` to distinguish "credentials not provisioned yet" from a real DB failure and act accordingly (e.g. call `EnsureCredentials` vs retry).

nadheesh: Also cleaned up the same pattern in `provisionCredentials` — replaced the compound `err != nil && !errors.Is(err, gorm.ErrRecordNotFound)` with explicit nested branching so the not-found vs real-DB-error distinction is immediately clear.

nadheesh: fixed

menakaj (approved):

PR #960, opened 2026-05-25
Add a runtime catalog for AMP instrumentation versions

### LINKED ISSUE #962: Add the capability to add new instrumentation versions without upgrading AMP
Today an AMP install can only use the `amp-instrumentation` versions that were available when that release of AMP was built. When we publish a newer instrumentation version, a customer can't pick it from the create-agent form until they upgrade their whole AMP install.

We want the supported set to be configurable on a running install. Operators should be able to add a new version through normal config, have it appear in the Console, and let their agent developers use it. The same path should work for air-gapped installs where images come from an internal registry.

RAVEENSR:
## Purpose

Resolves https://github.com/wso2/agent-manager/issues/962

A new `amp-instrumentation` version becomes selectable on an AMP install only after that install upgrades the whole AMP product. Today three places freeze the supported set at product build time: the Console TS constant `SUPPORTED_INSTRUMENTATION_VERSIONS`, the `agent-manager-service` env `OTEL_SUPPORTED_INSTRUMENTATION_VERSIONS`, and the platform default env. So when AMP publishes a new instrumentation version a month after a customer installed, that customer can't pick it without a product upgrade. This contradicts the redesign decision that the instrumentation version is independent of the AMP product version.

## Goals

Move the supported-version set out of build-time TS / env values into a runtime catalog the server owns. The catalog ships with a bundled baseline embedded into the binary, and a platform admin can extend it through Helm values without rebuilding the product. The Console reads it over an API instead of importing TS constants, so adding a version is one `helm upgrade` away rather than a product release away.

## Approach

There's a new `instrumentation` Go package that owns the catalog. It embeds `baseline.json` (generated from `.github/release-config.json` by a Make target) and reads an optional YAML extension file at `/etc/amp/instrumentation-extension.yaml`. Effective set is the union, deduped by version, extension wins on collision. The default version (still set via `OTEL_DEFAULT_INSTRUMENTATION_VERSION`) is validated against the effective set at startup; the old `OTEL_SUPPORTED_INSTRUMENTATION_VERSIONS` env is removed entirely.

The server exposes the catalog (plus the Python set, derived from `utils.Buildpacks`) at:

```
GET /api/v1/orgs/{orgName}/agent-build-options
```

Response carries `python.{defaultVersion,supportedVersions}` and `instrumentation.{defaultVersion,versions[].pythonVersions}`. Wire DI threads the catalog into a new `AgentBuildOptionsController`; the existing `validateInstrumentationVersion` path in `services/agent_manager.go` switches from the deleted config field to `instrumentation.GetCatalog()`. A cross-field validator was added too: agent create and update now reject a (python, instrumentation) pair that isn't compatible per the catalog entry, mirroring the Console-side filter. The pair check is gated on the agent's effective auto-instrumentation setting on both paths — agents with auto-instrumentation off never pull an init-container image, so the check is skipped.

The Helm chart grows two values under `agentManagerService.config.otel`: `defaultInstrumentationVersion` (existed only as an env before, now a real chart value) and `additionalInstrumentationVersions` (the extension list). The list renders into a ConfigMap mounted at `/etc/amp/`, and the Deployment gets a `checksum/instrumentation-extension` annotation so `helm upgrade` actually rolls the pod when the list changes.

The Console drops the hardcoded `SUPPORTED_INSTRUMENTATION_VERSIONS` / `DEFAULT_INSTRUMENTATION_VERSION` / `SUPPORTED_PYTHON_VERSIONS` / `DEFAULT_PYTHON_VERSION` constants, fetches via a new `useAgentBuildOptions` React Query hook, and adopts a Python-first form flow: pick Python, then the instrumentation dropdown filters to entries whose `pythonVersions` cover that Python. When the filter is empty the form renders an info message ("No AMP-provided instrumentation is available for Python X.Y. You can deploy this agent and instrument it manually.") and stays submittable with a null `instrumentationVersion`. The pip-install snippet for Docker agents now uses the fetched default instead of the deleted constant. The seed effect also normalises a stale `languageVersion` that disappears from the supported set across a build-options refetch, so an open form doesn't carry an unsupported value into submit.

**Boot-time guards.** Three checks run at server start so misconfigurations fail at `helm install` time, not at first agent-create:

- The default instrumentation version must be present in the effective catalog.
- The default instrumentation's `pythonVersions` must overlap the buildpack-supported Python set (otherwise the create-agent form would offer no valid combination).
- The hardcoded default Python version (`3.11`) must be present in `utils.Buildpacks` — pruning it without updating the default would otherwise silently advertise an unsupported value.

## User stories

Platform admin running an internal-registry install wants a new instrumentation version available to their agent developers without a product upgrade. They mirror the init-container image to their registry, add an entry under `additionalInstrumentationVersions` pointing `imageRepository` at the mirror, and `helm upgrade`. The new version shows up in the Console.

Platform admin running a connected install wants the same flow without the mirror step. Same Helm value, `imageRepository` left at the public `ghcr.io` path, `helm upgrade`. Done.

Agent developer wants to pick the instrumentation version that works with their framework. The form shows what's actually available on this install, filtered by their chosen Python version. If nothing covers their Python, they get a clear message and can still deploy with manual instrumentation.

## Release note

Operators can now extend the supported `amp-instrumentation` version set per install through a new Helm value (`agentManagerService.config.otel.additionalInstrumentationVersions`). The create-agent form populates its version dropdowns from a new runtime endpoint instead of compiled-in constants, so a new version becomes selectable after a `helm upgrade` rather than a full product upgrade.

## Documentation

A new operator-facing page covers the catalog end to end: bundled baseline vs. operator extension, the Helm value with an example, the platform default and its startup validation, internal-registry usage, and the failure-mode table. See `documentation/docs/administration/instrumentation-catalog.mdx`. The existing user-facing instrumentation page (`documentation/docs/components/amp-instrumentation.mdx`) was updated to frame its version-mapping table as the bundled baseline and to link operators to the new page. The sidebar gets a new `Administration` category for this and future operator-task content.

The maintainer runbook at `python-instrumentation-provider/RELEASING.md` was updated to reflect the new model: bumping the platform default now goes through the Helm chart value and requires regenerating `baseline.json`; the old "add to Console options" steps are removed since the Console reads the catalog at runtime.

## Automation tests

- Unit tests
  - `agent-manager-service/instrumentation/catalog_test.go`: baseline-only Load, default-not-in-set, extension adds new version, extension overrides bundled, missing file is silent, malformed YAML errors, missing required fields error. Plus `All()`/`Get()` return defensive copies and `SetCatalog(nil)` panics at the source.
  - `agent-manager-service/services/agent_manager_test.go`: `validateInstrumentationVersion` reads from the catalog; cross-field `validatePythonInstrumentationPair` accepts compatible pairs and rejects incompatible ones, naming both fields in the error; `validateEffectivePythonInstrumentationPair` falls back to the catalog default when the request pin is absent; `buildpackPythonVersion` normalises patch suffixes, whitespace, capital-P languages and similar edge cases.
  - `agent-manager-service/utils/buildpack_types_test.go`: `SupportedPythonVersions()` returns the bare-minor list parsed from the Python entry in `Buildpacks`.
  - `agent-manager-service/controllers/agent_build_options_controller_test.go`: response shape, both Python and instrumentation blocks populated, plus `compareVersionsDesc` correctness on the multi-digit minor case (`0.10.0` vs `0.2.1`) so the API's newest-first ordering stays semver-correct.

- CI consistency guard
  - `agent-manager-service/instrumentation/helm_consistency_test.go`: reads `values.yaml`, asserts `defaultInstrumentationVersion` exists in `baseline.json`. Runs with `go test ./...`; catches the most common release-day mistake (bumping the chart default without regenerating the embedded baseline) at PR time instead of at the customer's `helm install`.

- Integration tests: N/A in this PR. Existing API integration tests still exercise the agent-create validation path; the change is intentionally narrow (one validator function now reads from the catalog, one new cross-field validator was added) so existing coverage continues to apply.

- Manual smoke: tested end-to-end on the local docker-compose stack. Mounted an extension YAML with a `0.99.0` entry, hard-refreshed the create-agent form, confirmed `0.99.0` appears in the instrumentation dropdown alongside `0.2.1`. Then removed the mount and confirmed the dropdown reverts to baseline-only.

## Security checks

- Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines: yes
- Ran FindSecurityBugs plugin and verified report: yes
- Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets: yes

## Samples

N/A — the change is platform configuration, not a feature with sample agents.

## Related PRs

None.

## Migrations (if applicable)

N/A. The catalog lives entirely in process memory (embedded baseline + optional file mount); nothing persists to the database. No new migration was added. Existing migrations on `agent_configs` (017 / 018) already carry the per-agent `instrumentation_version` column and aren't touched.

## Test environment

- Go 1.25
- macOS 25.5.0 (Darwin) and Linux (Docker Desktop)
- Postgres 16 (the version `docker-compose` pulls)
- Helm 3, k3d-based render checks
- Node 22 for the Console build
- Chrome (latest) and Firefox (latest) for the Console smoke

## Learning

Weighed four directions for distributing instrumentation versions to running installs: a remote catalog only, a customer-extensible config only, a hybrid of the two with a bundled baseline, and deriving the set from `ghcr` tag listings. The hybrid (this PR) wins because internal-registry installs need a path that doesn't depend on outbound HTTP, and a tag-listing approach carries no compatibility metadata (the `traceloop-sdk` pin, the supported Python set). An optional remote-catalog poll is a natural follow-on but doesn't unblock the customer pain on its own — the bundled + extension floor does.

The wiring uses Google Wire DI for the new controller and a package-level `instrumentation.GetCatalog()` accessor for the legacy validator boundary in `services/agent_manager.go` — same pattern as the existing `config.GetConfig()`. Doing proper DI for the legacy path would have required threading the catalog through three or four constructors that don't otherwise need it; the package accessor is the smaller, lower-blast-radius change.

## Release Notes

* **New Features**
  * Added agent build options API endpoint providing dynamic Python and instrumentation version selection.
  * Instrumentation versions now determined by server-side catalog (bundled baseline plus admin extensions) instead of hardcoded values.
  * Console now fetches and displays compatible instrumentation versions based on selected Python version.

* **Bug Fixes**
  * Enhanced Python and instrumentation version compatibility validation during agent creation and updates.


agent-manager-service/api/agent_build_options_routes.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"net/http"
+
+	"github.com/wso2/agent-manager/agent-manager-service/controllers"
+	"github.com/wso2/agent-manager/agent-manager-service/middleware"
+)
+
+func registerAgentBuildOptionsRoutes(mux *http.ServeMux, ctrl controllers.AgentBuildOptionsController) {
+	middleware.HandleFuncWithValidation(mux, "GET /orgs/{orgName}/agent-build-options", ctrl.GetAgentBuildOptions)
```
hanzjk, line 27: Is this path added to the OpenAPI spec? Response and Request types are autogenrated based on api spec. (run `make spec`)

agent-manager-service/controllers/agent_build_options_controller.go
```diff
@@ -0,0 +1,142 @@
...
+	versions := c.catalog.All()
+	// Newest-first by numeric-component compare. Lexicographic sort
+	// inverts "0.10.0" vs "0.2.1" once the catalog grows past 0.9.x.
+	sorted := make([]instrumentation.Version, len(versions))
+	copy(sorted, versions)
+	sort.Slice(sorted, func(i, j int) bool {
+		return compareVersionsDesc(sorted[i].Version, sorted[j].Version)
+	})
+
+	resp := agentBuildOptionsResponse{
```
hanzjk: We can use the auto-generated response types from api spec.

agent-manager-service/api/agent_build_options_routes.go
RAVEENSR: Fixed in 65948f1c. Added `/orgs/{orgName}/agent-build-options` to `docs/api_v1_openapi.yaml`, ran `make spec`, and the new files under `agent-manager-service/spec/` carry the generated path + types.

agent-manager-service/controllers/agent_build_options_controller.go
RAVEENSR: Fixed in 65948f1c. Controller now uses `spec.AgentBuildOptionsResponse` (and the nested `AgentBuildOptionsPython` / `AgentBuildOptionsInstrumentation` / `AgentBuildOptionsInstrumentationEntry`) instead of the inline DTOs. JSON wire shape is unchanged.

hanzjk (approved):

PR #961, opened 2026-05-25
Implement MCP Proxy support

Oshanath:
## Summary by CodeRabbit

* **New Features**
  * Full MCP proxy management: create, list, view, update, delete, and server-info fetch
  * Agent-level MCP server bindings: add, list, update, remove, and per-agent MCP URL injection
  * Policy catalog & editor: browse policies, edit parameters with schema-driven UI, and validate
  * Automatic deployment orchestration to gateways and MCP lifecycle events broadcasting
  * UI pages and hooks for MCP proxies and agent bindings in the console

[Review Change Stack](https://app.coderabbit.ai/change-stack/wso2/agent-manager/pull/961?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


agent-manager-service/controllers/mcp_proxy_controller.go
```diff
@@ -0,0 +1,276 @@
...
+	log.Info("CreateMCPProxy: starting", "orgName", orgName)
+
+	var req spec.MCPProxy
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		log.Error("CreateMCPProxy: failed to decode request", "error", err)
+		utils.WriteErrorResponse(w, http.StatusBadRequest, "Invalid request body")
+		return
+	}
+
+	resp, err := c.mcpProxyService.Create(ctx, orgName, "system", &req)
```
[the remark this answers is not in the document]
Oshanath: This is the same behaviour for llm proxies as well. When a better approach is identified, all places will be fixed in bulk. Hence can ignore this for now.

agent-manager-service/repositories/mcp_proxy_repository.go
```diff
@@ -0,0 +1,211 @@
...
+			return err
+		}
+		return r.artifactRepo.Delete(tx, proxy.UUID.String())
+	})
+}
+
+// Count counts MCP proxies for an organization.
+func (r *MCPProxyRepo) Count(orgUUID string) (int, error) {
+	return r.artifactRepo.CountByKindAndOrg(models.KindMCPProxy, orgUUID)
+}
```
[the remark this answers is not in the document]
Oshanath: But this method is for getting org level mcp proxy count. mcp proxy mappings have a separate repository file. So this count is correct.

console/workspaces/pages/mcp-proxies/src/subComponents/ViewMCPProxy.tsx
```diff
@@ -0,0 +1,1098 @@
...
+              {policies.map((policy, index) => (
+                <Box
+                  key={`${policy.name}-${policy.version}-${index}`}
+                  draggable
+                  onDragStart={() => setDraggedPolicyIndex(index)}
+                  onDragEnd={() => setDraggedPolicyIndex(null)}
+                  onDragOver={(event) => event.preventDefault()}
+                  onDrop={() => handleDropPolicy(index)}
+                  onClick={() => setEditingPolicyIndex(index)}
+                  sx={{
```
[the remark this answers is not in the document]
Oshanath: Rest of the UI is the same. We can ignore this for now.

agent-manager-service/services/mcp_proxy_service.go
```diff
@@ -0,0 +1,1350 @@
...
+	}
+
+	upstream := req.Upstream
+	if upstream.Main != nil {
+		auth, err := s.prepareMCPUpstreamAuthForStorage(nil, upstream.Main.Auth)
+		if err != nil {
+			return nil, err
+		}
+		upstream.Main.Auth = auth
+	}
```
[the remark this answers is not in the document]
Oshanath: Environments isn't implemented yet in agent manager. Environments are custom.

agent-manager-service/services/mcp_proxy_service.go
```diff
@@ -0,0 +1,1350 @@
...
+
+	exists, err := s.repo.Exists(handle, orgUUID)
```
menakaj: # Non-Atomic Exists + Create Race Condition

```go
  // PROBLEMATIC - outside any transaction
  exists, err := s.repo.Exists(handle, orgUUID)
  if exists {
      return nil, utils.ErrMCPProxyExists
  }
  // ... window here allows duplicate creation ...
  if err := s.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
      return s.repo.Create(tx, proxy, handle, name, version, orgUUID)
  }); err != nil { ... }
```
  Two concurrent CreateMCPProxy calls for the same (handle, orgUUID) both pass the Exists check and both race to insert. The DB migration (023_create_mcp_proxies.go) adds no unique constraint on mcp_proxies. Only artifacts has any unique enforcement (via FK), but that's not checked.

  Fix: Either add a UNIQUE (handle, org_name) constraint to artifacts (or mcp_proxies via the artifact FK), and handle the constraint error in repo.Create with ON CONFLICT DO NOTHING, or move the existence check inside the same transaction.

agent-manager-service/services/mcp_proxy_service.go
```diff
@@ -0,0 +1,1350 @@
...
+	if req.Upstream.Main == nil || strings.TrimSpace(req.Upstream.Main.URL) == "" {
+		return nil, utils.ErrInvalidInput
+	}
+	if err := validateMCPProxyUpstreamURLs(ctx, req.Upstream); err != nil {
+		return nil, fmt.Errorf("%w: %w", utils.ErrInvalidURL, err)
+	}
+
+	upstream := req.Upstream
+
+	proxy, err := s.repo.GetByHandle(handle, orgUUID)
```
menakaj: # Read-Modify-Write Outside Transaction in Update

```go
proxy, err := s.repo.GetByHandle(handle, orgUUID)  // READ outside tx
  // ... mutate proxy fields ...
  if err := s.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
      return s.repo.Update(tx, proxy, orgUUID)        // WRITE inside tx
  }); err != nil { ... }
```

  Between the outer GetByHandle and the transaction's Update, another request can modify the same proxy. The Update uses a WHERE on UUID with no optimistic locking, so it silently overwrites concurrent changes.

  Fix: Move GetByHandle inside the transaction. Add a version/updated_at column and use optimistic locking (WHERE updated_at = ?) or a SELECT FOR UPDATE.

agent-manager-service/services/mcp_proxy_deployment.go
```diff
@@ -0,0 +1,656 @@
...
+	}
+
+	gateways, err := s.resolveSelectedMCPProxyGateways(orgName, gatewayIDs)
+	if err != nil {
+		return err
+	}
+	return s.deployMCPProxyToGateways(proxy, orgName, gateways)
+}
+
+func (s *MCPProxyService) resolveSelectedMCPProxyGateways(orgName string, gatewayIDs []string) ([]*models.Gateway, error) {
```
menakaj: Every deployment function — autoDeployToActiveGateways, deployMCPProxyToSelectedGateways, redeployMCPProxyToCurrentGateways, deployMCPProxyToGateways, deployMCPProxyToGateway, DeployAgentMCPMapping, redeployAgentMCPMappingsForProxy — takes no context.Context parameter. These all perform DB writes and NATS broadcasts, none of which can be cancelled on request timeout or shutdown.

  This is compounded by three explicit _ = ctx suppressions in the service layer:
  - MCPProxyService.Get 
  - MCPProxyService.Delete 
  - MCPProxyService.AgentMCPServerURLs

  Fix: Thread ctx context.Context as the first parameter through the entire deploy call chain. Pass it to all repository and NATS calls.

agent-manager-service/services/mcp_proxy_service.go
```diff
@@ -0,0 +1,1350 @@
...
+		if err := s.deployMCPProxyToSelectedGateways(created, orgUUID, req.Gateways); err != nil {
+			return convertModelMCPProxyToSpec(created), fmt.Errorf("proxy created but deployment failed: %w", err)
+		}
+	}
+
+	return convertModelMCPProxyToSpec(created), nil
+}
+
+// List retrieves MCP proxies for an organization.
+func (s *MCPProxyService) List(orgUUID string, limit, offset int) (*models.MCPProxyListResponse, error) {
```
menakaj:
```go
func (s *MCPProxyService) List(orgUUID string, limit, offset int) (*models.MCPProxyListResponse, error) {
  func (s *MCPProxyService) ListAvailableMCPPolicies(orgUUID string) (*models.MCPPolicyAvailabilityResponse, error) {
```

  Both perform DB queries with no context, violating the project's own CLAUDE.md rule: "Every method that performs I/O must accept context.Context as its first parameter." Callers (the controllers) have a live
  request context that should propagate for timeouts/cancellation.

agent-manager-service/services/gateway_internal_service.go
```diff
@@ -92,13 +97,70 @@ func NewGatewayInternalAPIService(
...
+// GetActiveMCPProxyDeploymentByGateway retrieves the currently deployed MCP proxy artifact.
+func (s *GatewayInternalAPIService) GetActiveMCPProxyDeploymentByGateway(ctx context.Context, proxyID, orgName, gatewayID string) (map[string]string, error) {
+	proxy, err := s.mcpProxyRepo.GetByUUID(proxyID, orgName)
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return s.getActiveMCPDeploymentByArtifact(ctx, proxyID, orgName, gatewayID)
+		}
+		return nil, fmt.Errorf("failed to get MCP proxy: %w", err)
+	}
+	if proxy == nil {
```
menakaj:
```go  
proxy, err := s.mcpProxyRepo.GetByUUID(proxyID, orgName)
  if err != nil {
      if errors.Is(err, gorm.ErrRecordNotFound) {
          return s.getActiveMCPDeploymentByArtifact(...)  // fallback
      }
      return nil, fmt.Errorf("failed to get MCP proxy: %w", err)
  }
  if proxy == nil {  // dead code: GORM never returns nil,nil on First()
      return s.getActiveMCPDeploymentByArtifact(...)
  }
```

  This is correct only for the non-nil proxy path. However, the proxy == nil fallback is dead code (gorm.First either populates the value or returns ErrRecordNotFound), which suggests the intent was more liberal
  fallback. The comment-less redundancy here is a latent bug waiting for someone to add GetByUUID semantics that return nil, nil.

  Fix: Remove the dead nil check. Add a comment documenting the UUID-vs-artifact-ID dual lookup behavior.

agent-manager-service/services/mcp_proxy_service.go
Oshanath: This exists check is just a pre-check. The actual checking happens in the `Create` function in `mcp_proxy_repository.go`. The artifact creation attempt is made first, and if the handle-org unique constraint  is violated, the mcp proxy creation does not happen. This flow is identical to that of llm providers.

CLAUDE.md
```diff
@@ -0,0 +1,3 @@
+Read these files before taking any action in this repository:
```
jhivandb: Shall we make this a AGENTS.md aswell? claude-code natively reads those too, and it also works with codex,cursor etc...

agent-manager-service/services/mcp_proxy_service.go
Oshanath: Fixed

agent-manager-service/services/mcp_proxy_deployment.go
Oshanath: Fixed

agent-manager-service/services/mcp_proxy_service.go
Oshanath: Fixed

agent-manager-service/services/gateway_internal_service.go
Oshanath: Fixed

CLAUDE.md
Oshanath: This local file was added as a mistake. Removed it.

menakaj (approved):

PR #985, opened 2026-06-01
Surface typed CLIErrors from auth.Login for known failures

### LINKED ISSUE #786: amctl: Surface typed clierr.CLIError from auth.Login instead of always wrapping to Transport
## Summary

Currently, `internal/amctl/auth/login.go` returns plain `fmt.Errorf`-wrapped errors on all failure paths (e.g. lines 77, 133). As a result, `cmd/login.go` cannot distinguish known auth failures (invalid_grant, unauthorized, user-cancelled PKCE, etc.) from generic transport errors, and always wraps them as `clierr.Transport`.

## Work to do

1. **Introduce typed `clierr.CLIError` returns in `internal/amctl/auth/login.go`** for known auth failure cases, including at minimum:
   - `invalid_grant` / token exchange failures
   - Unauthorized (401) responses from the token endpoint
   - User-cancelled PKCE browser flow

2. **Update `internal/amctl/cmd/login.go`** to pass through typed `clierr.CLIError` from `auth.Login` when present, instead of always wrapping to `clierr.Transport`:
   ```go
   if err != nil {
       var ce clierr.CLIError
       if errors.As(err, &ce) {
           return render.Error(opts.IO, scope, ce)
       }
       return render.Error(opts.IO, scope, clierr.Newf(clierr.Transport, "%v", err))
   }
   ```

## Context

- Identified during review of PR #753 (comment: https://github.com/wso2/agent-manager/pull/753#discussion_r3169955209)
- Deferred from PR #753 because both halves must land together to be meaningful.

/cc @jhivandb

nimeshk03:
## Purpose
`auth.Login` returned plain `fmt.Errorf`-wrapped errors on all failure paths, so `cmd/login.go` could not distinguish a known auth failure from a transport error — both became `CLI_TRANSPORT`.

Resolves #786

## Goals
- Map HTTP 401 and `invalid_grant` token-endpoint errors to `clierr.Unauthorized`
- Map user-cancelled PKCE flow (context cancellation, stdin EOF) to a new `clierr.AuthLoginCancelled` code
- Pass typed `clierr.CLIError` values through `cmd/login.go` unchanged instead of always wrapping as `CLI_TRANSPORT`

## Approach
- Add `classifyTokenError` in `auth/login.go` that inspects `*oauth2.RetrieveError` and returns a typed `CLIError` for known cases; returns `nil` for anything else so the existing `fmt.Errorf` wrap still applies
- In `loginPKCE`, check for `context.Canceled`, `context.DeadlineExceeded`, and `io.EOF` before calling `classifyTokenError`
- In `cmd/login.go`, use `errors.As` to pass through a `clierr.CLIError` from `auth.Login` before falling back to the `CLI_TRANSPORT` wrap
- Add `AuthLoginCancelled = "AUTH_LOGIN_CANCELLED"` constant to `clierr`

## User stories
As a script author using `amctl login`, I want the JSON error `code` field to distinguish auth rejections from transport failures so I can handle each case without parsing message strings.

## Release note
`amctl login` now emits `UNAUTHORIZED` for 401/`invalid_grant` errors and `AUTH_LOGIN_CANCELLED` when the browser login is cancelled, instead of always emitting `CLI_TRANSPORT`.

## Documentation
N/A — internal CLI error code change, no user-facing doc impact.

## Training
N/A

## Certification
N/A

## Marketing
N/A

## Automation tests
- Unit tests
  - `pkg/auth/login_test.go` (new): 8 table-driven cases for `classifyTokenError` covering 401, `invalid_grant`, wrapped errors, 403, unknown codes, plain errors, nil response
  - `pkg/cmd/login_test.go` (new): 4 cases verifying typed `CLIError` passthrough, plain errors become `CLI_TRANSPORT`, cancelled login passes `AuthLoginCancelled`, missing `--url` returns `INVALID_FLAG`
- Integration tests
  - N/A

## Security checks
- Ran FindSecurityBugs plugin and verified report? N/A (Go CLI)

## Samples
N/A

## Related PRs
#753 — original `amctl` CLI PR where this fix was deferred

## Migrations (if applicable)
N/A

## Test environment
Ubuntu 26.04, Go 1.26.0

## Learning
N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved classification and clearer JSON error codes/messages for OAuth/login failures (unauthorized, invalid credentials).
  * Browser-based login cancellations are now detected and reported with a dedicated error code.
  * Authentication failures are rendered consistently in command output.

* **Tests**
  * Added unit and command tests validating the new login error classifications and JSON error output.

### DIFF
--- a/cli/pkg/auth/login.go
+++ b/cli/pkg/auth/login.go
@@ -18,13 +18,17 @@ package auth
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io"
+	"net/http"
 	"strings"
 
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/clientcredentials"
 
 	"github.com/wso2/agent-manager/cli/pkg/browser"
+	"github.com/wso2/agent-manager/cli/pkg/clierr"
 	"github.com/wso2/agent-manager/cli/pkg/clients"
 	"github.com/wso2/agent-manager/cli/pkg/config"
 	"github.com/wso2/agent-manager/cli/pkg/iostreams"
@@ -74,6 +78,9 @@ func loginClientCredentials(ctx context.Context, opts LoginOptions) (*config.Ins
 	}
 	tok, err := cc.Token(ctx)
 	if err != nil {
+		if ce := classifyTokenError(err); ce != nil {
+			return nil, ce
+		}
 		return nil, fmt.Errorf("client_credentials token exchange: %w", err)
 	}
 
@@ -90,6 +97,22 @@ func loginClientCredentials(ctx context.Context, opts LoginOptions) (*config.Ins
 			Scopes:       scopes,
 		},
 	}, nil
+}
+
+// classifyTokenError maps oauth2 token-endpoint errors to typed CLIErrors.
+// Returns nil for errors that don't have a known classification.
+func classifyTokenError(err error) error {
+	var re *oauth2.RetrieveError
+	if !errors.As(err, &re) {
+		return nil
+	}
+	if re.Response != nil && re.Response.StatusCode == http.StatusUnauthorized {
+		return clierr.New(clierr.Unauthorized, "token endpoint rejected the request (401)")
+	}
+	if re.ErrorCode == "invalid_grant" {
+		return clierr.Newf(clierr.Unauthorized, "invalid_grant: %s", re.ErrorDescription)
+	}
+	return nil
 }
 
 func loginPKCE(ctx context.Context, opts LoginOptions) (*config.Instance, error) {
@@ -131,6 +154,12 @@ func loginPKCE(ctx context.Context, opts LoginOptions) (*config.Instance, error)
 
 	tok, err := authCodePKCE(ctx, oauthCfg, opts.IO, openBrowser)
 	if err != nil {
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) || errors.Is(err, io.EOF) {
+			return nil, clierr.New(clierr.AuthLoginCancelled, "browser login cancelled")
+		}
+		if ce := classifyTokenError(err); ce != nil {
+			return nil, ce
+		}
 		return nil, fmt.Errorf("authorization code exchange: %w", err)
 	}
 
--- a/cli/pkg/auth/login.go
+++ b/cli/pkg/auth/login.go
@@ -28,8 +28,8 @@ import (
 	"golang.org/x/oauth2/clientcredentials"
 
 	"github.com/wso2/agent-manager/cli/pkg/browser"
-	"github.com/wso2/agent-manager/cli/pkg/clierr"
 	"github.com/wso2/agent-manager/cli/pkg/clients"
+	"github.com/wso2/agent-manager/cli/pkg/clierr"
 	"github.com/wso2/agent-manager/cli/pkg/config"
 	"github.com/wso2/agent-manager/cli/pkg/iostreams"
 )
--- a/cli/pkg/auth/login.go
+++ b/cli/pkg/auth/login.go
@@ -78,10 +78,7 @@ func loginClientCredentials(ctx context.Context, opts LoginOptions) (*config.Ins
 	}
 	tok, err := cc.Token(ctx)
 	if err != nil {
-		if ce := classifyTokenError(err); ce != nil {
-			return nil, ce
-		}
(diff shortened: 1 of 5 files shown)


cli/pkg/cmd/login.go
```diff
@@ -104,6 +105,10 @@ func runLogin(ctx context.Context, opts *LoginOptions) error {
 		OpenBrowser:  opts.OpenBrowser,
 	})
 	if err != nil {
+		var ce clierr.CLIError
+		if errors.As(err, &ce) {
+			return render.Error(opts.IO, scope, ce)
+		}
```
jhivandb: **`render.Error` already does this.** `asCLIError` (`render.go:107-120`) and `textError` (`render.go:86-89`) both run `errors.As(err, &cliErr)` — typed `CLIError`s pass through with their code, everything else falls back to `Transport`. This block plus the new `errors` import collapses to `return render.Error(opts.IO, scope, err)` with identical behavior.

cli/pkg/auth/login.go
```diff
@@ -74,6 +78,9 @@ func loginClientCredentials(ctx context.Context, opts LoginOptions) (*config.Ins
 	}
 	tok, err := cc.Token(ctx)
 	if err != nil {
+		if ce := classifyTokenError(err); ce != nil {
+			return nil, ce
+		}
 		return nil, fmt.Errorf("client_credentials token exchange: %w", err)
```
jhivandb: The classify-then-`fmt.Errorf` pattern is duplicated in `loginPKCE` (line 160). Could fold into one `wrapTokenError(err, ctx)` helper so each site becomes a single `return nil, wrapTokenError(err, "client_credentials")`.

cli/pkg/cmd/login_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"errors"
+	"testing"
+
+	"github.com/wso2/agent-manager/cli/pkg/auth"
+	"github.com/wso2/agent-manager/cli/pkg/clierr"
+	"github.com/wso2/agent-manager/cli/pkg/config"
+	"github.com/wso2/agent-manager/cli/pkg/iostreams"
+)
+
+func TestRunLogin_TypedErrorPassthrough(t *testing.T) {
```
jhivandb: These 4 tests each repeat ~13 lines of setup + envelope-decode + code assertion. `auth/login_test.go` in this same PR is already table-driven — a table here would cut the duplication and match that style.

jhivandb (commented):
/simplify cleanup notes

cli/pkg/cmd/login.go
nimeshk03: Addressed in the latest commit. Removed the explicit CLIError handling from runLogin and delegated error rendering directly to render.Error(...).

cli/pkg/auth/login.go
nimeshk03: Addressed in the latest commit. Extracted the classification-and-wrapping logic into wrapTokenError() and updated both login flows to use the helper. The PKCE path retains its existing fallback error context.

cli/pkg/cmd/login_test.go
nimeshk03: Addressed in the latest commit. I consolidated the repeated login error test cases into a single table-driven test to reduce duplication and keep the assertions in one place.

nimeshk03:
Hi @jhivandb,

Thank you for the review and feedback.

I've addressed all comments in the latest commit.

Let me know if you have any further feedback or if there are any additional changes required.

jhivandb:
Hi @nimeshk03, I don't think we should merge this right now.

The existing token exchange paths already preserve the OAuth error text:

return nil, fmt.Errorf("client_credentials token exchange: %w", err)
return nil, fmt.Errorf("authorization code exchange: %w", err)

So this mainly changes the structured JSON code from `CLI_TRANSPORT` to `UNAUTHORIZED`. That would be useful if we had a stable automation path depending on `amctl login --json`, but client-
credentials support is still minimal and the upcoming authz changes are going to change this area anyway.

I also don't see a real browser-cancel path that reaches `AUTH_LOGIN_CANCELLED`. Closing the browser tab doesn't call the callback, and an OAuth error redirect comes through as `authorization
error: ...`, not `context.Canceled`, `DeadlineExceeded`, or `io.EOF`.

Given that, I don't see enough value in adding new stable error codes here yet. I think we should defer this until the authz/client-credentials story is settled, then add the error contract
around the final flow.

Issue #1028, opened 2026-06-09
Improvements Related to RBAC Implementation


lashinijay:
### Description

The RBAC is implemented and enabled by default in the product with PR[1].

There are couple of improvements that needs to be done top of RBAC implementation.

- [x] Permissions of predefined roles should not be editable.
- [ ] Validate ouName path parameter with authorized token claims.

[1] - https://github.com/wso2/agent-manager/pull/914

lashinijay:
Related PR:

- https://github.com/wso2/agent-manager/pull/1053

PR #1034, opened 2026-06-09
Disabling modifying permissions of predefined roles

lashinijay:
## Related Issue

- https://github.com/wso2/agent-manager/issues/1028

## Summary by CodeRabbit

* **New Features**
  * Predefined roles are protected — they cannot be edited or deleted.
  * Permissions for predefined roles are read-only; modify actions and removal buttons disabled in the UI.
  * Role listings are scoped to your organizational unit and now exclude the Administrator role.
  * Roles now expose a read-only flag so the UI consistently prevents edits where applicable.
  * Role details and creation responses include the read-only indicator.

* **Chores**
  * Service configuration updated with the key manager JWKS endpoint.

### DIFF
--- a/console/workspaces/libs/types/src/api/identities.ts
+++ b/console/workspaces/libs/types/src/api/identities.ts
@@ -85,20 +85,13 @@ export interface UpdateGroupRequest {
 
 // --- Roles ---
 
-export const PREDEFINED_ROLES = [
-  "admin",
-  "developer",
-  "ai-lead",
-  "platform-engineer",
-  "Administrator",
-] as const;
-
 export interface ThunderRole {
   id: string;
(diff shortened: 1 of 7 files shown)


console/workspaces/libs/types/src/api/identities.ts
```diff
@@ -85,6 +85,14 @@ export interface UpdateGroupRequest {
 
 // --- Roles ---
 
+export const PREDEFINED_ROLES = [
```
menakaj: Can we get this from the backend instead of defining at ui and backend both places? 

ex: set readonly flag from backend for default roles or something similar?

menakaj (approved):

PR #1053, opened 2026-06-12
Add validation for ouName path parameter with authorized token claim

lashinijay:
## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary by CodeRabbit

* **Security & Bug Fixes**
  * Added early organization-path validation across many HTTP endpoints so requests with missing or mismatched org in the URL are rejected immediately, preventing further processing.
  * Tightened authorization to treat empty organization identifiers as invalid, returning forbidden responses when org context is missing or does not match.


agent-manager-service/controllers/helpers.go
```diff
@@ -0,0 +1,39 @@
...
+
+// validateOrgFromPath validates that the org in the path matches the caller's token org.
+func validateOrgFromPath(w http.ResponseWriter, ctx context.Context, pathOrg string) bool {
+	resolvedOrg, ok := middleware.GetResolvedOrg(ctx)
+	if !ok {
+		utils.WriteErrorResponse(w, http.StatusForbidden, "missing org context")
+		return false
+	}
+	if pathOrg != resolvedOrg.OuHandle {
+		utils.WriteErrorResponse(w, http.StatusForbidden, "invalid organization access")
```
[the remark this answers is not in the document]
lashinijay: ouHandle is now enforced in the middleware. Since the JWT claims are system controlled, these two claims will always be available in the token

agent-manager-service/controllers/agent_configuration_controller.go
```diff
@@ -140,6 +144,10 @@ func (c *agentConfigurationController) GetAgentModelConfig(w http.ResponseWriter
 	agentName := r.PathValue(utils.PathParamAgentName)
 	configID := r.PathValue(utils.PathParamConfigId)
 
+	if !validateOrgFromPath(w, ctx, orgName) {
```
jhivandb: Shouldn't these be validated in the authz middleware? We're trying to keep validations out of the controllers right?

agent-manager-service/controllers/agent_configuration_controller.go
jhivandb, line 138: Shouldn't this use GetResolvedOrg and in the middleware we should check the path param against the orgHandle in the token?

lashinijay: what we do here is validate whether the ouName in path parameter is same as the ouName in the authorized token. If only those two matches the operation will continue.

jhivandb: I'm suggesting moving this validation to the middleware aswell, the path org should always match the org handle in the ouHandle right?

lashinijay: already addressed https://github.com/wso2/agent-manager/pull/1053/changes#diff-69ce9fe57fcb65b28bfb06accbabc70cee2ecf95ece9ca893d1589f89700f842R67-R72

RAVEENSR (approved):

PR #1143, opened 2026-06-22
Implement CLI for agent mcp proxy configs.

Oshanath:
# Add CLI commands for agent MCP configs

Adds `amctl agent mcp` commands to list, view, set, and unset per-environment MCP proxy bindings on an agent. This mirrors the existing `agent llm` command group (#1109), applying the LLM→MCP correspondence:

- **LLM provider → MCP proxy** (org-level handle bound per environment)
- **LLM proxy mapping → MCP proxy mapping** (the agent-level, per-environment configuration these commands manage)

## Examples

```bash
amctl agent mcp set order-bot --name primary --env dev --proxy github --url-env MCP_URL --apikey-env MCP_API_KEY
amctl agent mcp list order-bot
amctl agent mcp get order-bot --name primary
amctl agent mcp unset order-bot --name primary --env dev
```

## What's added

A new `cli/pkg/cmd/agent/mcp` package with four subcommands, registered under `agent`:

| Command | Behavior |
|---------|----------|
| `list`  | Lists the agent's configurations, filtered to MCP-typed configs. |
| `get`   | Shows a single MCP config, including its per-environment proxy bindings (env / proxy / url / status). |
| `set`   | Creates or updates an MCP proxy binding for one environment. Optional `--url-env` / `--apikey-env` map the proxy URL and API key to custom env var names; `--description` sets a description. |
| `unset` | Without `--env`, deletes the whole MCP config. With `--env`, removes only that environment's binding and preserves the rest. |

All commands support `--json` output and the global `--org` / `--project` flags, and the agent argument is optional (falls back to the configured default agent).

## Implementation notes

- **Dedicated endpoints**: uses the generated `*AgentMCPConfig*` client methods (the `/agents/{agent}/mcp-configs` path) for list/get/create/update/delete, rather than the LLM `/model-configs` endpoints. Both share the same request/response types.
- **Proxy binding field**: `--proxy` is written to `EnvModelConfigRequest.ProxyName`, the canonical field the server's `resolveEnvConfigProxyName` resolves first. On read-back, `proxyNameFromConfig` extracts the handle from the response (the server mirrors it across `providerName`/`proxyName`/`mcpProxyName`).
- **Type filtering**: configs are created with type `mcp` and every read filters on `type == "mcp"`, so LLM and MCP configs don't leak across the two command groups.
- **Read-merge-write**: `set` (env update) and `unset --env` re-send the full existing env-mapping set, because `PUT` replaces the entire `envMappings` map — updating one environment never drops its siblings.
- **Client-side validation**: `--url-env` / `--apikey-env` names are validated with `cmdutil.ValidateEnvVarName` before any API call, giving early feedback instead of a server rejection. MCP mappings use the same `url` / `apikey` env-var keys as LLM mappings.

## Tests

`go test ./...` (from `cli/`). New table/HTTP-stub tests cover each subcommand: human + JSON output, the MCP-type filter, the read-merge-write sibling-preservation guarantee, last-binding rejection, env-var name validation, and transport/server error paths.

## Summary by CodeRabbit

* **New Features**
  * Added MCP proxy configuration management to the agent CLI command with four new subcommands: `list` to view configurations, `get` to display specific settings, `set` to configure proxies per environment, and `unset` to remove configurations.

* **Documentation**
  * Updated build-from-source instructions to include proper archive extraction step.

### DIFF
--- a/cli/pkg/cmd/agent/mcp/unset.go
+++ b/cli/pkg/cmd/agent/mcp/unset.go
@@ -1,0 +1,170 @@
+// Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+//
+// WSO2 LLC. licenses this file to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package mcp
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/spf13/cobra"
+
+	openapi_types "github.com/oapi-codegen/runtime/types"
+
+	amsvc "github.com/wso2/agent-manager/cli/pkg/clients/amsvc/gen"
+	"github.com/wso2/agent-manager/cli/pkg/clierr"
+	"github.com/wso2/agent-manager/cli/pkg/cmdutil"
+	"github.com/wso2/agent-manager/cli/pkg/iostreams"
+	"github.com/wso2/agent-manager/cli/pkg/render"
+)
+
+type UnsetOptions struct {
+	IO           *iostreams.IOStreams
+	Client       func(context.Context) (*amsvc.ClientWithResponses, error)
+	ResolveScope func(*cobra.Command, bool, bool) (string, string, error)
+	MakeScope    func(org, proj, agent string) render.Scope
+	ResolveAgent func([]string) (string, []string, error)
+
+	Org       string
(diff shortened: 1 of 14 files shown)


cli/pkg/cmd/agent/mcp/unset.go
```diff
@@ -0,0 +1,170 @@
...
+	id, found, err := findMCPConfigByName(ctx, client, o.Org, o.Proj, o.AgentName, o.Name)
+	if err != nil {
+		return render.Error(o.IO, o.Scope, err)
+	}
+	if !found {
+		return render.Error(o.IO, o.Scope, clierr.Newf(clierr.NotFound, "no MCP configuration named %q for agent %q", o.Name, o.AgentName))
+	}
+
+	if o.Env == "" {
+		return runUnsetDelete(ctx, o, client, id)
```
jhivandb, line 125: Lets add a prompt if unset is called without specifying an environment. Could wipe a prod binding when you were intending on only changing dev

Oshanath: This was implemented with the latest commit.

jhivandb (approved):

PR #1195, opened 2026-06-26
Agent config api key management

Oshanath:
## Summary

Adds API key management support for LLM providers, MCP proxies, and external agent configurations.

## Changes

- Added backend API key management endpoints for:
  - LLM providers
  - MCP proxies
  - External agent model configs
  - External agent MCP configs
- Updated OpenAPI specs and regenerated service/CLI clients.
- Added API key purpose migration for user-managed and console-managed keys.
- Revokes API keys when API key security is removed.
- Added shared console API key management components.
- Added API key management UI sections/tabs for:
  - LLM provider security
  - MCP proxy security
  - Configure-agent LLM proxy details
  - Configure-agent MCP server details
- Improved LLM provider and MCP proxy security tab UX.
- Cleaned up MCP proxy UI issues in the add form and table.

## Summary by CodeRabbit

* **New Features**
  * Added masked API key listing plus create/rotate/revoke for external-agent MCP and LLM configuration environments (scoped by config + environment).
  * Added masked API key listing for LLM providers and LLM proxies, and expanded MCP proxy API key management and MCP proxy/policy management endpoints.
  * Introduced new UI tabs/sections for MCP proxy, LLM provider, and external-agent configuration API key management.
* **Bug Fixes**
  * API key auth enablement/disable behavior is more consistent, including revoking user-managed keys when API-key authentication is turned off.
* **Security & Permissions**
  * Expanded OAuth/RBAC scopes for MCP server API key management; API key location now supports header/query only.


agent-manager-service/services/llm_provider_apikey_service.go
```diff
@@ -48,6 +51,48 @@ func NewLLMProviderAPIKeyService(
...
+
+	stored, err := s.apiKeyRepo.ListByArtifact(providerID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list API keys: %w", err)
+	}
+
+	keys := make([]models.APIKeyInfo, 0, len(stored))
+	for _, k := range stored {
+		// Only surface user-managed keys; hide console-managed and test keys.
+		if k.Purpose != models.APIKeyPurposeUserManaged {
```
jhivandb: 🤖 _Agent review — reuse/dedup._

This `stored → []APIKeyInfo` loop (filter `APIKeyPurposeUserManaged`, copy Name/DisplayName/MaskedAPIKey/Status, RFC3339-format CreatedAt/ExpiresAt) is duplicated **byte-for-byte in four places**:

- `services/llm_provider_apikey_service.go:73-91` (here)
- `services/llm_proxy_apikey_service.go:73-91`
- `services/mcp_proxy_service.go:459-477`
- `services/agent_configuration_service.go:457-475`

**Suggestion:** extract one helper in `models/apikey.go`, e.g. `func ToUserManagedAPIKeyInfos(stored []StoredAPIKey) []APIKeyInfo`, since both `StoredAPIKey` and `APIKeyInfo` live in `models`. Each call site then collapses to `return &models.ListAPIKeysResponse{Keys: models.ToUserManagedAPIKeyInfos(stored)}, nil`, centralizing the "hide non-user-managed" rule and the timestamp formatting (~75 duplicated lines removed).

agent-manager-service/services/agent_configuration_service.go
```diff
@@ -362,6 +391,237 @@ func (s *agentConfigurationService) revokeAllMCPMappingAPIKeys(ctx context.Conte
...
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return uuid.Nil, utils.ErrAgentConfigNotFound
+		}
+		return uuid.Nil, fmt.Errorf("failed to get agent configuration: %w", err)
+	}
+	if config == nil || config.AgentID != agentName || config.ProjectName != projectName {
+		return uuid.Nil, utils.ErrAgentConfigNotFound
+	}
+
+	envs, err := s.infraResourceManager.ListOrgEnvironments(ctx, orgName)
```
jhivandb, line 435: 🤖 _Agent review — reuse/dedup._

`resolveMCPMappingArtifactUUID` (lines 401-425) and `resolveLLMProxyHandleForConfig` (lines 532-556) share an identical ~25-line prologue: `GetByUUID` + `ErrRecordNotFound` mapping, the `config == nil || AgentID != agentName || ProjectName != projectName` ownership check, then `ListOrgEnvironments` + the linear scan mapping `envName → envUUID`. Only the final lookup differs (MCP mapping artifact vs. LLM proxy handle).

**Suggestion:** extract `resolveConfigAndEnvUUID(ctx, orgName, projectName, agentName, configUUID, envName) (*AgentConfiguration, string /*envUUID*/, error)` and have both resolvers call it, each keeping only its distinct trailing lookup.

agent-manager-service/services/agent_configuration_service.go
```diff
@@ -362,6 +391,237 @@ func (s *agentConfigurationService) revokeAllMCPMappingAPIKeys(ctx context.Conte
...
+		if env.Name == envName {
+			envUUID = env.UUID
+			break
+		}
+	}
+	if envUUID == "" {
+		return uuid.Nil, utils.ErrAgentConfigNotFound
+	}
+
+	mappings, err := s.envMCPMappingRepo.ListByConfig(ctx, config.UUID)
```
jhivandb: 🤖 _Agent review — efficiency._

Redundant DB query. `GetByUUID` (line 401) already `Preload("EnvMCPMappings").Preload("EnvMCPMappings.Artifact")` (see `repositories/agent_configuration_repository.go:97-100`), so this extra `envMCPMappingRepo.ListByConfig(...)` re-fetches data already loaded on `config`.

**Suggestion:** iterate the preloaded `config.EnvMCPMappings` and read `mapping.ArtifactUUID` directly — exactly as the sibling `resolveLLMProxyHandleForConfig` does with `config.EnvMappings` (line 558). Drops one query per List/Create/Rotate/Revoke call.

agent-manager-service/controllers/agent_configuration_controller.go
```diff
@@ -849,3 +857,293 @@ func getStringPtr(s string) *string {
...
+		return
+	}
+
+	response, err := c.agentConfigService.CreateMCPConfigAPIKey(ctx, orgName, projName, agentName, configUUID, envName, &models.CreateAPIKeyRequest{
+		Name:        name,
+		DisplayName: displayName,
+		ExpiresAt:   specReq.ExpiresAt,
+	})
+	if err != nil {
+		switch {
```
jhivandb: 🤖 _Agent review — reuse/dedup._

Two repeated patterns across the six new key handlers:

1. **Error switch** — `switch { case ErrAgentConfigNotFound → 404; case ErrGatewayNotFound → 503; default → log + 500 }` is copy-pasted in `CreateMCPConfigAPIKey`, `RotateMCPConfigAPIKey`, `RevokeMCPConfigAPIKey`, `CreateLLMConfigAPIKey`, `RotateLLMConfigAPIKey`, `RevokeLLMConfigAPIKey`. A `writeConfigAPIKeyError(w, log, op, err)` helper (mirroring the existing `writeMCPProxyAPIKeyError` precedent) would collapse all six.

2. **Name/displayName unwrap** (lines 915-926) — the `*specReq.Name`/`*specReq.DisplayName` deref + `if name == "" && displayName == ""` validation is identical in `CreateMCPConfigAPIKey` and `CreateLLMConfigAPIKey`. Extract a small helper that yields `(name, displayName string, ok bool)`.

jhivandb (commented):
🤖 **Automated review by an agent (Claude Code `/simplify`)** — backend Go only; console/UI and generated code skipped. These are **quality/simplification** observations (reuse, dedup, efficiency), *not* correctness-bug findings. Four inline comments below.

agent-manager-service/services/agent_configuration_service.go
```diff
@@ -362,6 +391,237 @@ func (s *agentConfigurationService) revokeAllMCPMappingAPIKeys(ctx context.Conte
...
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return "", utils.ErrAgentConfigNotFound
+		}
+		return "", fmt.Errorf("failed to get agent configuration: %w", err)
+	}
+	if config == nil || config.AgentID != agentName || config.ProjectName != projectName {
+		return "", utils.ErrAgentConfigNotFound
+	}
+
+	envs, err := s.infraResourceManager.ListOrgEnvironments(ctx, orgName)
```
jhivandb: 🤖 _Agent review — efficiency (larger scope)._

Both resolvers translate a single env **name → UUID** by calling `infraResourceManager.ListOrgEnvironments(ctx, orgName)` and linear-scanning the result. `ListOrgEnvironments` makes **two remote OpenChoreo calls** (`GetOrganization` + `ListEnvironments`) and returns the whole org env list — paid on every List/Create/Rotate/Revoke key operation, just to find one env.

A targeted single-env lookup already exists on the OpenChoreo client: `ocClient.GetEnvironment(ctx, orgName, envName)` (used e.g. `services/agent_manager.go:607`, `services/agent_token_manager.go:271`). It isn't currently exposed on the `infraResourceManager` interface (`services/infra_resource_manager.go:33`), so this is **larger scope** — it means adding `GetEnvironment` to that interface — but it would turn a list-all-and-scan into a direct lookup on every key op.

agent-manager-service/services/agent_configuration_service.go
```diff
@@ -362,6 +391,237 @@ func (s *agentConfigurationService) revokeAllMCPMappingAPIKeys(ctx context.Conte
...
+	}
+	return uuid.Nil, utils.ErrAgentConfigNotFound
+}
+
+// ListMCPConfigAPIKeys returns the masked, user-managed API key(s) for an external
+// agent's MCP configuration in the given environment.
+func (s *agentConfigurationService) ListMCPConfigAPIKeys(
+	ctx context.Context, orgName, projectName, agentName string, configUUID uuid.UUID, envName string,
+) (*models.ListAPIKeysResponse, error) {
+	if s.llmProxyAPIKeyService == nil || s.llmProxyAPIKeyService.broadcaster.apiKeyRepo == nil {
```
jhivandb: 🤖 _Agent review — altitude (larger scope)._

The per-config MCP key handlers reach generic key plumbing by tunneling through a sibling service's **unexported** field: `s.llmProxyAPIKeyService.broadcaster.apiKeyRepo.ListByArtifact(...)` and `s.llmProxyAPIKeyService.broadcaster.broadcastCreate/Rotate/Revoke(...)` (also at lines 322, 332, 357, 360, 371, 374, 452, 492, 508, 523).

MCP-config key management isn't really an "LLM proxy" operation — this `x.y.z.method` train-wreck is a sign the shared capability is owned by the wrong object. `apiKeyBroadcaster` already owns `apiKeyRepo` + all the broadcast methods and is the seam every key family routes through.

**Suggestion (larger scope — touches wiring):** inject the `apiKeyBroadcaster` (or a small `APIKeyManager` wrapping repo + gateways) directly into `agentConfigurationService` instead of borrowing it transitively through `llmProxyAPIKeyService`. Requires a `wire_gen.go` regen, hence flagged separately from the quick cleanups.

jhivandb (commented):
🤖 **Automated review by an agent (Claude Code) — follow-up.** Two deeper, larger-scope suggestions (altitude/efficiency). Unlike the previous batch these touch code beyond the immediate diff (an interface addition / a wiring change), so treat them as design notes rather than quick cleanups.

agent-manager-service/db_migrations/migration_list.go
```diff
@@ -16,7 +16,7 @@
 
 package dbmigrations
 
-const latestVersion = 25
+const latestVersion = 26
```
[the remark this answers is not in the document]
Oshanath: There is currently no way to know which permanent API keys were created by users or created by the console for internal usage. So the used approach is to consider all llm and mcp keys are console managed. This is fine since old user managed keys were not manageable, and they will continue to work until expiry. Only newly created api keys will be manageable, as only new keys are discernible as user managed and console managed.

agent-manager-service/services/agent_configuration_service.go
Oshanath: This is too large a change to tackle in this PR. This will need to be addressed separately.

agent-manager-service/services/agent_configuration_service.go
Oshanath: This is too large a change to tackle in this PR. This will need to be addressed separately.

agent-manager-service/services/agent_configuration_service.go
jhivandb: Its valid tho, shall we create issues to track this?

agent-manager-service/services/agent_configuration_service.go
jhivandb: Its valid tho, shall we create issues to track this?

Oshanath: This will be fixed with the next push.

agent-manager-service/services/agent_configuration_service.go
Oshanath: Issue created: https://github.com/wso2/agent-manager/issues/1207

agent-manager-service/controllers/agent_configuration_controller.go
Oshanath: Fixed.

agent-manager-service/services/agent_configuration_service.go
Oshanath: Fixed

agent-manager-service/services/agent_configuration_service.go
Oshanath: Fixed

agent-manager-service/services/llm_provider_apikey_service.go
Oshanath: Fixed

console/workspaces/pages/mcp-proxies/src/subComponents/MCPProxySecurityTab.tsx
```diff
@@ -181,52 +181,54 @@ export function MCPProxySecurityTab({
...
+                sx={{
+                  "& .MuiInputBase-input": {
+                    fontFamily: "monospace",
+                  },
+                }}
+              />
+            </FormControl>
+          </Grid>
+        </Grid>
+      )}
```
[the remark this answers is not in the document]
Oshanath: @coderabbitai check now

agent-manager-service/services/llm_provider_apikey_service.go
```diff
@@ -48,6 +51,48 @@ func NewLLMProviderAPIKeyService(
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get LLM provider: %w", err)
+	}
+	if provider == nil {
+		return nil, utils.ErrLLMProviderNotFound
+	}
+
+	stored, err := s.apiKeyRepo.ListByArtifact(providerID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list API keys: %w", err)
```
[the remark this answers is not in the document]
Oshanath: @coderabbitai check now

agent-manager-service/services/llm_proxy_apikey_service.go
```diff
@@ -48,6 +51,48 @@ func NewLLMProxyAPIKeyService(
...
+	proxy, err := s.proxyRepo.GetByID(proxyID, orgID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get LLM proxy: %w", err)
+	}
+	if proxy == nil {
+		return nil, utils.ErrLLMProxyNotFound
+	}
+
+	stored, err := s.apiKeyRepo.ListByArtifact(proxy.UUID.String())
+	if err != nil {
```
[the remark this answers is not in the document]
Oshanath: @coderabbitai check now

agent-manager-service/services/mcp_proxy_service.go
```diff
@@ -427,6 +443,42 @@ func (s *MCPProxyService) Delete(ctx context.Context, orgUUID, proxyID string) e
...
+// The plain key value is never returned — only the masked representation.
+func (s *MCPProxyService) ListAPIKeys(ctx context.Context, orgUUID, proxyID string) (*models.ListAPIKeysResponse, error) {
+	proxy, err := s.getMCPProxyByID(ctx, orgUUID, proxyID)
+	if err != nil {
+		return nil, err
+	}
+
+	stored, err := s.apiKeyBroadcaster.apiKeyRepo.ListByArtifact(proxy.UUID.String())
+	if err != nil {
+		return nil, fmt.Errorf("failed to list API keys: %w", err)
```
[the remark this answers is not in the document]
Oshanath: @coderabbitai check now

agent-manager-service/controllers/agent_configuration_controller.go
```diff
@@ -849,3 +859,260 @@ func getStringPtr(s string) *string {
...
+	}
+
+	response, err := c.agentConfigService.ListMCPConfigAPIKeys(ctx, orgName, projName, agentName, configUUID, envName)
+	if err != nil {
+		if errors.Is(err, utils.ErrAgentConfigNotFound) {
+			utils.WriteErrorResponse(w, http.StatusNotFound, "Configuration not found")
+			return
+		}
+		log.Error("ListMCPConfigAPIKeys: failed to list API keys", "error", err)
+		utils.WriteErrorResponse(w, http.StatusInternalServerError, "Failed to list API keys")
```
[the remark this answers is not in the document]
Oshanath: The concern assumes `ListMCPConfigAPIKeys` and `ListLLMConfigAPIKeys` can return `ErrAgentConfigNotExternal` and `ErrGatewayNotFound`, but neither error is reachable on the list code paths. `ErrAgentConfigNotExternal` is produced in exactly one place — `ensureExternalAgentForAPIKey` — which only the create/rotate/revoke service methods call; the two list methods deliberately skip that check, so the 403 branch can never fire. `ErrGatewayNotFound` is only raised when broadcasting key changes to gateways (broadcaster/platform-gateway code), and the list methods never broadcast — they only do DB reads (`ListByArtifact`) plus OpenChoreo org/env/project lookups via the resolve helpers, none of which touch gateway code. So mapping only `ErrAgentConfigNotFound` is exactly correct for the errors a list can actually produce, and mirroring the create/rotate/revoke handlers would just add two dead branches.

agent-manager-service/controllers/mcp_proxy_apikey_controller.go
```diff
@@ -0,0 +1,167 @@
...
+	log := logger.GetLogger(ctx)
+	orgName := r.PathValue(utils.PathParamOrgName)
+	proxyID := r.PathValue(utils.PathParamProxyId)
+
+	var specReq spec.CreateLLMAPIKeyRequest
+	if err := json.NewDecoder(r.Body).Decode(&specReq); err != nil {
+		log.Error("CreateMCPProxyAPIKey: failed to decode request", "orgName", orgName, "proxyID", proxyID, "error", err)
+		utils.WriteErrorResponse(w, http.StatusBadRequest, "Invalid request body")
+		return
+	}
```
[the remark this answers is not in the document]
Oshanath: Lenient decoding is the Go stdlib default and the codebase-wide convention here — all 23 request-decoding controllers use plain json.NewDecoder(r.Body).Decode(...), and the only DisallowUnknownFields in the repo is in generated client code agent-manager-service/spec/client.go#L574, not hand-written handlers. There's also no correctness/security impact for these endpoints: the decoded CreateLLMAPIKeyRequest is explicitly remapped field-by-field into the internal models.CreateAPIKeyRequest (Name/DisplayName/ExpiresAt at mcp_proxy_apikey_controller.go:95-99, so unknown fields can't cause mass-assignment, and trailing JSON after the first value is simply ignored by Decode. Bolting strict parsing onto just these two handlers would create a lone inconsistency; if strict decoding is wanted, it should be a deliberate repo-wide policy (a shared decode helper), not a one-controller patch.

menakaj (dismissed):

menakaj (approved):

PR #1214, opened 2026-06-30
feat: Introduce environment-specific identity providers (env-ThunderID)

yashed:
This PR introduces support for **Environment-Specific Identity Providers (env-Thunder)**. It establishes the backend APIs, frontend UI, Helm configurations, and setup/removal scripts to automatically provision, track, and deprovision isolated identity providers alongside environment creations and deletions.

### Backend Architecture (`agent-manager-service`)
* Created naming, slugification, and validation helpers in `clients/thundersvc/naming.go` to keep release names, namespaces, and OIDC issuer hosts within safe character lengths (<= 53 characters for Helm release names, <= 63 characters for DNS labels) using collision-safe SHA-256 prefixes.
* Implemented `ListThunderInstances` in `environmentService`. It queries environment objects from OpenChoreo, gates them by checking the local database (`gateway_environment_mappings`), and formats OIDC issuer, token, JWKS, and namespace details.
* Added corresponding controller endpoints, API routes, and schema data models.
### Frontend Console (`console`)
* Introduced a new **"Thunder Instances"** view in the organization section (`console/workspaces/pages/gateways/src/ThunderInstances.Organization.tsx`) allowing operators to list, copy endpoint URLs, and locate credentials.
* Registered new client API definitions, React hooks, route mappings, and sidebar navigation items.
### Helm Charts & Provisioning Scripts (`deployments`)
* Modified `wso2-amp-thunder-extension` to configure OIDC CORS mappings, support HTTPRoutes, and mount a custom CA bundle (Mozilla trust store + platform Root CA) to ensure HTTPS trust chain compatibility.
* Created `add-environment-thunder.sh` and `remove-environment-thunder.sh` to handle Helm deployment, credential mirroring to OpenBao (failing explicitly if Bao is unreachable), and namespace-bound cleanup.
* Updated `add-environment.sh` and `remove-environment.sh` to download these helper scripts to temporary files and run them with validation checks, resolving shell command injection risks.

## User stories
* **As a Platform Developer**, I want a dedicated identity provider (OIDC) per environment, so that agent OAuth credentials are fully isolated between staging, production, and development environments.
* **As a Developer**, I want to view all active OIDC endpoints (token, JWKS, issuer) and copy credentials for my environment in a single unified Console view.

## Automation tests
 - Unit tests 
   > Do Unit Testing
 - Integration tests
   > N/A

## Security checks
 - Ran FindSecurityBugs plugin and verified report? no

 

## Summary by CodeRabbit

* **New Features**
  * Added per-environment Thunder ID provisioning/deprovisioning with deterministic issuer, token, and JWKS endpoint details surfaced by the platform.
  * Introduced Identity Providers console pages to list Thunder instances and view an instance’s read-only OAuth2/infra endpoints (with copy-to-clipboard).
  * Updated setup to provision the default environment’s Thunder instance earlier and display console URLs/credentials when available.
* **Bug Fixes**
  * Improved Thunder instance discovery by probing reachability and excluding missing/unnamed/unreachable environments.
  * Enhanced env-Thunder wildcard routing and certificate/SAN coverage to prevent endpoint/TLS mismatches.


deployments/scripts/add-environment.sh
```diff
@@ -160,7 +232,48 @@ else
...
+if [ "${PROVISION_THUNDER:-true}" = "true" ]; then
+    echo ""
+    echo "🔐 Provisioning Thunder ID instance for '${ENV_NAME}'..."
+    # THUNDER_SCRIPT_URL can be set by the caller to pin a specific release ref.
+    # The console sets it via getRawScriptUrl; the default follows main (latest).
+    SCRIPT_BASE_URL="${SCRIPT_BASE_URL:-https://raw.githubusercontent.com/wso2/agent-manager/main/deployments/scripts}"
+    THUNDER_SCRIPT_URL="${THUNDER_SCRIPT_URL:-${SCRIPT_BASE_URL}/add-environment-thunder.sh}"
+    script_tmp="$(mktemp)"
+    if curl -fsSL "${THUNDER_SCRIPT_URL}" -o "$script_tmp"; then
+      if ENV_NAME="${ENV_NAME}" DISPLAY_NAME="${DISPLAY_NAME}" ORG_NAME="${ORG_NAME}" \
```
jhivandb, line 267: The chained `bash "$script_tmp"` inherits the caller's `CHART_VERSION`. The console drawer sets `CHART_VERSION` to the AMP release version (`CreateEnvironmentDrawer.tsx:96` -> `getAmpVersionHelm()`), which `add-environment.sh` correctly uses as the *gateway-extension* OCI chart version (line 304). But `add-environment-thunder.sh` reuses the same var for a *different* chart -- the upstream ThunderID OCI chart (default `oci://ghcr.io/thunder-id/helm-charts/thunderid`, line 530; consumed at 557 and as the image tag at 614). So every console-driven creation does `helm pull .../thunderid --version <AMP release>` (e.g. 0.16.0), which doesn't exist (validated ThunderID version is 0.45.0) -> provisioning fails -> silent fallback to platform Thunder. The explicit env passthrough at 253-254 doesn't include `CHART_VERSION`, but prefixing only *adds* vars; it doesn't sandbox. Fix: `CHART_VERSION="" bash "$script_tmp"` (or a dedicated `THUNDER_CHART_VERSION`).

deployments/scripts/add-environment.sh
```diff
@@ -172,27 +285,51 @@ RELEASE_NAME="api-platform-${ORG_NAME}-${ENV_NAME}"
...
+    #   internal_jwks = Thunder's K8s service DNS — avoids routing through the ingress
+    #                   Service name follows the chart template: {{ .Release.Name }}-service
+    THUNDER_RELEASE="$(_thunder_release_name "${ORG_NAME}" "${ENV_NAME}")"
+    THUNDER_ISSUER="http://$(_thunder_host "${ORG_NAME}" "${ENV_NAME}"):8080"
+    THUNDER_INTERNAL_JWKS="http://${THUNDER_RELEASE}-service.${THUNDER_RELEASE}.svc.cluster.local:8090/oauth2/jwks"
+    HELM_ARGS+=(
+        --set "apiGateway.config.policyConfigurations.jwtauth_v1.keymanagers[1].name=ThunderKeyManager"
+        --set "apiGateway.config.policyConfigurations.jwtauth_v1.keymanagers[1].issuer=${THUNDER_ISSUER}"
+        --set "apiGateway.config.policyConfigurations.jwtauth_v1.keymanagers[1].jwks.remote.uri=${THUNDER_INTERNAL_JWKS}"
+        --set "apiGateway.config.policyConfigurations.jwtauth_v1.keymanagers[1].jwks.remote.skipTlsVerify=true"
```
jhivandb, line 346: Helm `--set` replaces entire arrays rather than merging -- setting `keymanagers[0].*` (299-302) discards the chart's default 2-element list (`values.yaml:128-140`: `agent-manager-service` + `ThunderKeyManager`). In this `else` branch only `keymanagers[0]` is re-asserted, so `keymanagers[1]` (the platform-Thunder `ThunderKeyManager`) is dropped -- yet this log claims "gateway keeps the chart's default ThunderKeyManager", which a `helm template` render shows is false (single-element list). Net: every platform-Thunder JWT is rejected with 401. Either re-assert `keymanagers[1]` here (pointing at platform Thunder) or avoid indexed `--set` on this list.

deployments/scripts/remove-environment.sh
```diff
@@ -43,6 +45,29 @@ RELEASE_NAME=$(echo "$RELEASE_NAME" | head -c 53 | sed 's/-*$//')
 echo "=== Removing Environment: ${ENV_NAME} ==="
 echo ""
 
+# --- Step 0: Deprovision the environment's Thunder ID instance ---
```
jhivandb: Step 0 destroys the env-Thunder release + namespace (`helm uninstall` + `namespace delete` -- `remove-environment-thunder.sh:91-117`) *before* the AMS health check (72) and the DELETE call (88) that gate removal. If the token is expired or DELETE returns non-204, the environment stays live in AMS/OpenChoreo while its identity provider (Thunder DB, per-agent OAuth clients, JWKS endpoint) is already gone -- no restore path. Pre-PR, all Thunder teardown ran only after a successful 204/404. Move Step 0 below the `case` block.

deployments/scripts/add-environment.sh
```diff
@@ -172,27 +285,51 @@ RELEASE_NAME="api-platform-${ORG_NAME}-${ENV_NAME}"
...
+    --set agentManager.apiUrl="${AGENT_MANAGER_INTERNAL_API}"
+    --set apiGateway.controlPlane.host="${AGENT_MANAGER_INTERNAL_CP}"
+    --set apiGateway.controlPlane.tls.insecureSkipVerify=true
+    --set "apiGateway.config.policyConfigurations.jwtauth_v1.keymanagers[0].name=agent-manager-service"
+    --set "apiGateway.config.policyConfigurations.jwtauth_v1.keymanagers[0].issuer=agent-manager-service"
+    --set "apiGateway.config.policyConfigurations.jwtauth_v1.keymanagers[0].jwks.remote.uri=${AGENT_MANAGER_INTERNAL_JWKS}"
+    --set "apiGateway.config.policyConfigurations.jwtauth_v1.keymanagers[0].jwks.remote.skipTlsVerify=true"
+)
+if [ -n "$CHART_VERSION" ]; then
+    HELM_ARGS+=(--version "${CHART_VERSION}")
```
jhivandb, line 328: `THUNDER_PROVISIONED` reflects only this run's fetch/execute (243/257), so the advertised "re-run with the same ENV_NAME (idempotent)" is unsafe under transient failure. On a re-run where the curl at 252 blips, `THUNDER_PROVISIONED=false` -> the `else` branch runs `helm upgrade` (330) with no `--reuse-values` and without the `keymanagers[1]` sets (319-323) -> the gateway's JWKS reverts to platform Thunder while agents still mint tokens against the *live* env-Thunder issuer -> 401s. Persist provisioning state (a record/ConfigMap) so re-runs know env-Thunder exists.

deployments/quick-start/install.sh
```diff
@@ -1506,6 +1506,32 @@ fi
...
 
+# Provision per-environment Thunder instance for the default environment.
+# Mirrors setup-openchoreo.sh:install_default_env_thunder().
+# Must run after the platform Thunder TLS cert is issued (cert-manager) and
+# after the default environment exists (created by platform-resources above).
+log_info "Waiting for platform Thunder TLS certificate to be issued by cert-manager..."
+if kubectl wait --for=condition=Ready certificate/amp-thunder-extension-local-tls \
+        -n amp-thunder --timeout=120s 2>/dev/null; then
+    log_success "Platform Thunder TLS certificate is ready"
+else
```
jhivandb: Wrong namespace. `thunder-https-tls.yaml:59` hardcodes `Certificate {{ .Release.Name }}-local-tls` into `openchoreo-control-plane` (comment at :54 says so), but this `kubectl wait` targets `-n amp-thunder`. It fails instantly with NotFound (stderr suppressed via `2>/dev/null`), so there's no synchronization -- `add-environment-thunder.sh` can hard-fail at `platform_thunder_ca_cert` (:671) when cert-manager hasn't issued `amp-local-root-ca-secret` yet. `setup-openchoreo.sh:474` gets this right (`-n openchoreo-control-plane`, 300s). The 120s-vs-300s drift is itself a symptom of the duplicated block (see simplify comment at install.sh:1510).

agent-manager-service/services/environment_service.go
```diff
@@ -407,3 +409,42 @@ func toOCClientGatewayListenerSpec(l *models.GatewayListenerSpec) *occlient.Gate
...
+
+	instances := make([]models.ThunderInstanceResponse, 0, len(envs))
+	for _, env := range envs {
+		if env == nil || env.Name == "" {
+			continue
+		}
+
+		// Verify the environment has been provisioned (i.e. has gateway mappings in our DB).
+		// Environments without gateway mappings do not have a corresponding Thunder ID instance.
+		mappings, err := s.gatewayRepo.GetEnvironmentMappingsByEnvironmentID(env.UUID)
```
jhivandb: Existence is inferred only from gateway mappings existing, then endpoints are computed by naming convention (443-446) with no provisioning record or cluster check. Environments created with `PROVISION_THUNDER=false`, whose provisioning failed (non-fatal by design), or created pre-PR all have mappings but no Thunder -- yet the API (and the new console Identity page) advertise issuer/token/JWKS URLs that resolve to nothing. Gate on an actual provisioning record (or a live probe), not on `len(mappings) > 0`.

agent-manager-service/services/environment_service.go
```diff
@@ -407,3 +409,42 @@ func toOCClientGatewayListenerSpec(l *models.GatewayListenerSpec) *occlient.Gate
...
+			// Skip advertising Thunder for unprovisioned environments.
+			continue
+		}
+
+		instances = append(instances, models.ThunderInstanceResponse{
+			EnvName:      env.Name,
+			DisplayName:  env.DisplayName,
+			IsProduction: env.IsProduction,
+			IssuerURL:    thundersvc.ThunderIssuerURL(orgName, env.Name),
+			TokenURL:     thundersvc.ThunderTokenURL(orgName, env.Name),
```
jhivandb: `ThunderTokenURL`/`ThunderJWKSURL` (naming.go:96-98, 83-85) return cluster-internal `svc.cluster.local:8090` addresses -- naming.go:95 documents them as "reachable only inside the cluster". But this response is developer-facing (the model comment says developers use these to configure agents; the console Overview tab renders them with copy buttons). A developer outside the cluster copies the token endpoint -> DNS failure, even though a working external route exists via the HTTPRoute (`http://<org>-<env>.thunder.amp.localhost:8080/oauth2/...`, derivable from `ThunderHost`). Surface the external URL here.

deployments/scripts/add-environment-thunder.sh
```diff
@@ -0,0 +1,788 @@
...
+  local attempt
+  for attempt in 1 2 3; do
+    if ! curl -fsSL --connect-timeout 30 https://curl.se/ca/cacert.pem -o "$tmp_bundle" 2>/dev/null \
+        || ! grep -q "BEGIN CERTIFICATE" "$tmp_bundle"; then
+      rm -f "$tmp_bundle"
+      echo "❌ Could not fetch Mozilla CA bundle from https://curl.se/ca/cacert.pem"
+      echo "   Download it on a machine with internet access, then re-run:"
+      echo "     curl -fsSL https://curl.se/ca/cacert.pem -o /tmp/cacert.pem"
+      echo "     ENV_NAME=... bash $(basename "$0")"
+      exit 1
```
jhivandb: The Mozilla CA fetch exits fatally on the first curl failure -- the 3-attempt loop (:686) only retries *checksum* mismatches (:701), not download failures (:688 -> `exit 1` at :694). The printed recovery tells the operator to `curl ... -o /tmp/cacert.pem` (:692), but no code reads `/tmp/cacert.pem` (the script uses a `mktemp` path, :684) and there's no env-var override to point at a pre-downloaded bundle -- so the suggested fix can't work. It also re-downloads the ~230KB bundle on every run even when the target ConfigMap (:724) already exists. Add an env-var override + ConfigMap-existence short-circuit, and retry download failures (not just checksums).

deployments/quick-start/install.sh
```diff
@@ -1506,6 +1506,32 @@ fi
...
+fi
+
+log_info "Provisioning Thunder identity provider for the default environment..."
+ENV_THUNDER_PROVISIONED=false
+if ! install_default_env_thunder; then
+    log_warning "Default environment Thunder provisioning failed (non-fatal)"
+    echo "Re-run manually once the platform is ready:"
+    echo "  ENV_NAME=default DISPLAY_NAME=Default ORG_NAME=default CHART_VERSION=${VERSION} \\"
+    echo "  bash <(curl -fsSL https://raw.githubusercontent.com/wso2/agent-manager/amp/v${VERSION}/deployments/scripts/add-environment-thunder.sh)"
+else
```
jhivandb, line 1521: The recovery hint sets `CHART_VERSION=${VERSION}` (the AMP release, e.g. `0.0.0-dev`), but `add-environment-thunder.sh` interprets `CHART_VERSION` as the upstream ThunderID chart version (:557, image tag :614) -- so the copy-pasted command does `helm pull .../thunderid --version 0.0.0-dev`, which doesn't exist, and recovery can never succeed. `install-helpers.sh:244-246` and `setup-openchoreo.sh` both get this right by leaving `CHART_VERSION` unset (-> script default 0.45.0). Drop `CHART_VERSION=${VERSION}` from the hint.

deployments/scripts/remove-environment-thunder.sh
```diff
@@ -0,0 +1,135 @@
...
+  echo "  Environment: ${ENV_NAME}"
+  echo "  Release:     ${release}"
+  echo "  Namespace:   ${ns}"
+  echo ""
+}
+
+# Run main only when executed directly — not when sourced (e.g. by tests).
+if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
+  main "$@"
+fi
```
jhivandb, line 133: Under `set -euo pipefail` (:2), `${BASH_SOURCE[0]}` is unbound when the script is piped to bash (`curl ... | ENV_NAME=staging bash`) -- the exact invocation its own header (:10-12) and `remove-environment.sh`'s printed recovery (`remove-environment.sh:62`) document. It crashes with "BASH_SOURCE[0]: unbound variable" before `main` runs, so the env-Thunder release/namespace is never removed. The sibling `add-environment-thunder.sh:786` uses `${BASH_SOURCE[0]:-$0}` with a comment explaining exactly this hazard -- mirror it here.

agent-manager-service/clients/thundersvc/naming.go
```diff
@@ -0,0 +1,169 @@
...
+
+	thunderInternalPort = 8090
+	maxReleaseNameLen   = 53
+	truncatePrefixLen   = 46
+)
+
+// ThunderReleaseName returns the Helm release name (and namespace) for an env-Thunder instance.
+// Mirrors thunder_release_name() in add-environment-thunder.sh — must stay in sync.
+//
+// Format: amp-thunder-<org>-<env>, capped at 53 characters.
```
jhivandb, line 58: The release-name/host/namespace algorithm lives in four copies that must stay byte-for-byte in sync: here (`ThunderReleaseName`/`ThunderHost`), `add-environment-thunder.sh`'s `thunder_release_name`/`thunder_host`, `remove-environment-thunder.sh`'s `thunder_release_name`, and `add-environment.sh`'s `_thunder_release_name`/`_thunder_host`. A missed edit in any copy silently points the gateway/AMS at addresses where nothing is deployed. Deeper fix: make provisioning the single writer -- record issuer/JWKS/namespace at provision time (ConfigMap/OpenBao/AMS registration) and have Go + the remove script read that record instead of re-deriving. That also fixes correctness #6 and #7 at the root.

deployments/quick-start/install.sh
```diff
@@ -1506,6 +1506,32 @@ fi
 log_success "Platform Resources Extension installed successfully"
 echo ""
 
+# Provision per-environment Thunder instance for the default environment.
+# Mirrors setup-openchoreo.sh:install_default_env_thunder().
+# Must run after the platform Thunder TLS cert is issued (cert-manager) and
+# after the default environment exists (created by platform-resources above).
```
jhivandb: This wait + `install_default_env_thunder` block is duplicated from `setup-openchoreo.sh:466-489`, and the two have already drifted: this one waits in the wrong namespace (#5) and uses 120s vs 300s. `install_default_env_thunder` is also defined twice (here via `install-helpers.sh:247` and `setup-openchoreo.sh:443`). Factor one shared helper, or move the cert-wait into `add-environment-thunder.sh` itself so callers can't diverge.

deployments/quick-start/install.sh
```diff
@@ -1597,6 +1623,24 @@ if [[ "${SHOW_LOCALHOST_URLS:-true}" == "true" ]]; then
...
+
+# Print the default environment's Thunder ID console + admin credentials — the
+# release/namespace/host follow the fixed ENV_NAME=default ORG_NAME=default naming
+# convention (see thunder_release_name/thunder_namespace/thunder_host in
+# add-environment-thunder.sh). The password lives only in a K8s Secret (never written
+# to disk by add-environment-thunder.sh), so it's fetched fresh here rather than
+# re-printed from that script's own (long since scrolled-past) console output.
+if [[ "${ENV_THUNDER_PROVISIONED}" == "true" ]]; then
+  ENV_THUNDER_ADMIN_PASSWORD="$(kubectl get secret amp-thunder-default-default-admin-credentials \
+    -n amp-thunder-default-default -o jsonpath='{.data.password}' 2>/dev/null | base64 -d 2>/dev/null || true)"
```
jhivandb: `amp-thunder-default-default` / `...-admin-credentials` are hand-spelled derived names. For any org/env whose `amp-thunder-<org>-<env>` exceeds 53 chars, the real release name is hash-truncated (naming.go:50 / `thunder_release_name`), so this literal is wrong. Call the shared naming helper (or read the live release) instead.

deployments/setup/setup-openchoreo.sh
```diff
@@ -543,3 +598,32 @@ echo ""
...
+  [ -n "$_ns" ] || continue
+  _secret="${_ns}-admin-credentials"
+  kubectl get secret "$_secret" -n "$_ns" &>/dev/null 2>&1 || continue
+  _active_count=$((_active_count + 1))
+  if [ "$_active_count" -eq 1 ]; then
+    echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
+    echo "  Thunder ID — Provisioned Instances (save these credentials!)"
+    echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
+  fi
+  _host="$(kubectl get httproute "$_ns" -n openchoreo-control-plane -o jsonpath='{.spec.hostnames[0]}' 2>/dev/null || echo "")"
```
jhivandb: Reverse-deriving the host from the namespace (`${_ns#amp-thunder-}.thunder.amp.localhost`) is wrong for hash-truncated names: a truncated namespace is `amp-thunder-<prefix>-<hash>`, but `ThunderHost` truncates the `<org>-<env>` *label* at 56 chars + hash -- different prefix lengths, so this string isn't the real hostname. Prefer the HTTPRoute hostname (already fetched at :614) and drop the fallback, or compute via the shared helper.

agent-manager-service/clients/thundersvc/naming.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+const maxAgentAppNameLen = 100
+
+// AgentThunderAppName returns the OAuth app name to use in Thunder for a per-agent client.
+// Format: amp-agent-<org>-<project>-<agent>, truncated to 100 chars, trailing hyphen stripped.
+// The name mirrors amp-publisher-<org> but is fully scoped to project + agent to avoid collisions.
+func AgentThunderAppName(org, project, agent string) string {
+	org = slugify(org)
+	project = slugify(project)
```
jhivandb, line 289: `AgentThunderAppName` has no callers (verified by grep across the service). Same for `ThunderSystemClientSecretName` (:90) and the `thunderSystemClientID`/`thunderSystemClientSecretKey` constants (:28, :32) -- only `ThunderIssuerURL`/`TokenURL`/`JWKSURL`/`Namespace` are actually used. `slugify` (:154) is only reached via the dead `AgentThunderAppName`, and is a third slug helper alongside `services.slugify` (`evaluator_manager.go:325`) and `utils.GenerateCandidateName`. Remove the dead exports, or wire them up if they're intended for upcoming work.

console/workspaces/pages/env-thunders/src/subComponents/ViewThunderInstance.tsx
```diff
@@ -0,0 +1,141 @@
...
+              ))}
+            </Tabs>
+            <Divider />
+            <Box sx={{ p: 3 }}>
+              {tabIndex === 0 && (
+                <ThunderInstanceOverviewTab instance={instance} onCopy={handleCopy} />
+              )}
+              {tabIndex === 1 && <ThunderInstanceUsersTab />}
+              {tabIndex === 2 && <ThunderInstanceRolesTab />}
+              {tabIndex === 3 && <ThunderInstanceGroupsTab />}
```
jhivandb: `ThunderInstanceUsersTab`/`RolesTab`/`GroupsTab` are three ~26-line files that each just wrap `ThunderInstanceComingSoonTab` with a different icon+title (see `ThunderInstanceUsersTab.tsx`). Inline them directly in this tab switch -- three `<ThunderInstanceComingSoonTab illustration={...} title="..." />` lines -- and delete the files.

console/workspaces/pages/env-thunders/src/subComponents/ViewThunderInstance.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { ThunderInstanceUsersTab } from "./ThunderInstanceUsersTab";
+
+const TABS = ["Overview", "Users", "Roles", "Groups"] as const;
+
+export const ViewThunderInstance: React.FC = () => {
+  const { orgId, envName } = useParams<{ orgId: string; envName: string }>();
+  const [tabIndex, setTabIndex] = useState(0);
+  const [snackbarOpen, setSnackbarOpen] = useState(false);
+  const [snackbarMessage, setSnackbarMessage] = useState("");
+
```
jhivandb, line 44: Copy-feedback Snackbar state is hand-rolled here (47-48, 53-58, 131-136) and again in `ThunderInstancesTable.tsx:40-47`, and the two copies already diverge (this one carries a message, the table's doesn't). Use the shared `SnackBarProvider`/`useSnackBar` from `@agent-management-platform/views` (`SnackBar.tsx`) like pages/deploy and pages/mcp-proxies do.

agent-manager-service/services/environment_service.go
```diff
@@ -407,3 +409,42 @@ func toOCClientGatewayListenerSpec(l *models.GatewayListenerSpec) *occlient.Gate
...
+// ListThunderInstances returns the Thunder OAuth2 identity provider info for every
+// environment in the org that has been provisioned (i.e. has active gateway mappings).
+func (s *environmentService) ListThunderInstances(ctx context.Context, orgName string) (*models.ThunderInstanceListResponse, error) {
+	envs, err := s.ocClient.ListEnvironments(ctx, orgName)
+	if err != nil {
+		return nil, fmt.Errorf("list environments for org %s: %w", orgName, err)
+	}
+
+	instances := make([]models.ThunderInstanceResponse, 0, len(envs))
+	for _, env := range envs {
```
jhivandb: One sequential `GetEnvironmentMappingsByEnvironmentID` per environment (N+1) just to test `len() > 0`. A single batched `IN`/`EXISTS` query over all env UUIDs would replace the loop's existence check.

deployments/scripts/add-environment-thunder.sh
```diff
@@ -0,0 +1,788 @@
...
+    break
+  done
+  mozilla_bundle="$(cat "$tmp_bundle")"
+  rm -f "$tmp_bundle"
+  ca_bundle="${mozilla_bundle}
+${ca_pem}"
+
+  kubectl create namespace "$ns" --dry-run=client -o yaml | kubectl apply -f - >/dev/null 2>&1 || true
+
+  # Store the combined bundle as a ConfigMap.
```
jhivandb: Namespace create is duplicated -- :563 already ensures it exists; this `kubectl create namespace` is a guaranteed no-op (and silently `|| true`'d). Also, the Mozilla CA bundle (:687) is re-downloaded on every run even when the ConfigMap (:724) already exists -- overlaps #8.

jhivandb (commented):
AI Review: Some comments from claude

agent-manager-service/clients/thundersvc/naming.go
```diff
@@ -0,0 +1,239 @@
...
+
+	// 3. Fallback for Docker container development (macOS / host.docker.internal)
+	host := ThunderHost(org, env)
+	if probe("http://"+host+":8080/oauth2/jwks", "host.docker.internal:8080") {
+		return true
+	}
+
+	// 4. Fallback for Linux containers / host networking (127.0.0.1)
+	return probe("http://"+host+":8080/oauth2/jwks", "127.0.0.1:8080")
+}
```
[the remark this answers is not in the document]
yashed: Fixed by fanning out the per-environment `ThunderProbe` calls concurrently (`sync.WaitGroup`, same pattern as `agent_kind_service.go`) instead of probing sequentially, so latency no longer scales with environment count. `ThunderProbe` itself is unchanged — verified with existing tests + `-race`.

agent-manager-service/services/environment_service.go
```diff
@@ -407,3 +409,51 @@ func toOCClientGatewayListenerSpec(l *models.GatewayListenerSpec) *occlient.Gate
...
+		// The env-Thunder HTTPRoute routes http://<org>-<env>.thunder.amp.localhost:8080
+		// to the Thunder service, so all /oauth2/* paths are reachable via ThunderHost.
+		instances = append(instances, models.ThunderInstanceResponse{
+			EnvName:      env.Name,
+			DisplayName:  env.DisplayName,
+			IsProduction: env.IsProduction,
+			IssuerURL:    thundersvc.ThunderIssuerURL(orgName, env.Name),
+			TokenURL:     thundersvc.ThunderExternalTokenURL(orgName, env.Name),
+			JWKSURL:      thundersvc.ThunderExternalJWKSURL(orgName, env.Name),
+			Namespace:    thundersvc.ThunderNamespace(orgName, env.Name),
```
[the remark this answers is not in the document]
yashed: `ThunderProbe` is now wrapped behind an injectable `thundersvc.Prober` interface, wired through the constructor and DI (matching the `ocClient`/`clientmocks` pattern), with a generated `ThunderProberMock` and a dedicated "reachable" unit test covering the previously-untested success path. This comment is referencing a stale diff snapshot from before that commit.

agent-manager-service/services/environment_service.go
```diff
@@ -411,3 +416,67 @@ func toOCClientGatewayListenerSpec(l *models.GatewayListenerSpec) *occlient.Gate
...
+// ListThunderInstances returns the Thunder OAuth2 identity provider info for every
+// environment in the org whose env-Thunder instance is reachable.
+//
+// Reachability is determined by a live HTTP probe to the environment's JWKS endpoint,
+// NOT by inferring from gateway mappings. Gateway mappings only prove a gateway was
+// provisioned — not that Thunder was ever deployed (e.g. PROVISION_THUNDER=false,
+// a failed provision, or an environment created before this feature was added all have
+// mappings but no Thunder instance). Advertising dead endpoints would cause the console
+// Identity page to show broken issuer/token/JWKS URLs.
+func (s *environmentService) ListThunderInstances(ctx context.Context, orgName string) (*models.ThunderInstanceListResponse, error) {
```
jhivandb, line 429: **Architecture: Thunder existence is determined solely by live probing, with no persisted provisioning state.**

`ListThunderInstances` decides whether an environment has a Thunder IdP by probing its JWKS endpoint on every call (line 447), and the comment at line 458 concedes "reachability is the only reliable signal." Three problems follow from having no source of truth:

1. A transiently-down or restarting Thunder (probe exceeds the 2s timeout) silently disappears from the console — indistinguishable from "never provisioned."
2. `ThunderProbe` accepts any HTTP 200 with no JWKS-body validation, and its fallbacks target fixed local addresses, so an unprovisioned env can be reported as live with dead URLs (see the probe comment below).
3. Latency is bounded by the slowest probe (~8s worst case, per line 436) on a request that's hit on every page load.

Consider persisting a provisioning flag/record on the environment at provision time and treating the probe as a health signal layered on top, rather than as the existence oracle.

agent-manager-service/clients/thundersvc/naming.go
```diff
@@ -0,0 +1,239 @@
...
+
+// ThunderProbe checks whether an env-Thunder instance is reachable by performing a
+// HTTP GET to its JWKS endpoint. It tries the cluster-internal URL first, falling
+// back to the public URL, and finally host-header resolved fallbacks (e.g. host.docker.internal)
+// to support local development inside Docker containers. Callers treat a negative probe
+// as "not provisioned" and skip the env.
+//
+// A 2-second timeout avoids blocking the ListThunderInstances hot path on unreachable
+// instances while still being generous enough for a slow cold start.
+func ThunderProbe(ctx context.Context, org, env string) bool {
```
jhivandb, line 224: **Dev-machine network topology is baked into production service code.**

`ThunderProbe` tries four URLs sequentially at 2s each, and steps 3–4 are hardcoded to `host.docker.internal:8080` and `127.0.0.1:8080` with `Host`-header overrides. This chain is ungated by deployment mode, so in-cluster it burns timeout budget on fallbacks that can never succeed, and it's the source of the false-positive issue: any server on port 8080 that returns 200 regardless of `Host` (a dev proxy, another service) makes every environment probe as "reachable."

Consider resolving the probe target from configuration/an injected resolver, keeping dev overrides in config rather than a hardcoded fallback chain in the binary.

agent-manager-service/clients/thundersvc/naming.go
```diff
@@ -0,0 +1,239 @@
...
+// is not applied here (must match the un-collapsed bash implementation byte-for-byte).
+func ThunderHost(org, env string) string {
+	org = strings.ToLower(org)
+	env = strings.ToLower(env)
+	if org == "" || env == "" {
+		panic("org and env names must be valid alphanumeric slugs and not empty")
+	}
+	label := fmt.Sprintf("%s-%s", org, env)
+	if len(label) <= 63 {
+		return fmt.Sprintf("%s.thunder.amp.localhost", strings.TrimSuffix(label, "-"))
```
jhivandb: **The developer-facing Thunder URLs are compile-time constants, so non-localhost deployments get wrong issuer/token/JWKS URLs.**

`ThunderHost` / `ThunderIssuerURL` / `ThunderExternalTokenURL` hardcode `http://<label>.thunder.amp.localhost:8080`, and these values flow straight into the `ListThunderInstances` API response. The only Thunder config that exists (`THUNDER_BASE_URL`) is for the admin API and isn't consulted here, so any deployment with a custom domain, HTTPS, or a different gateway port returns issuer URLs that are wrong for every environment — fixable only by a code change and rebuild.

These should come from configuration (base domain / scheme / port), not constants.

deployments/scripts/add-environment.sh
```diff
@@ -160,7 +238,70 @@ else
...
+    # silently revert the gateway's JWKS back to platform Thunder while a live env-Thunder
+    # instance still exists and agents are minting tokens against it.
+    THUNDER_RELEASE_NAME="$(_thunder_release_name "${ORG_NAME}" "${ENV_NAME}")"
+    THUNDER_NS="${THUNDER_RELEASE_NAME}"
+    if helm status "${THUNDER_RELEASE_NAME}" --namespace "${THUNDER_NS}" > /dev/null 2>&1; then
+        echo "✅ Thunder ID instance already exists (Helm release: ${THUNDER_RELEASE_NAME})"
+        THUNDER_PROVISIONED=true
+    fi
+
+    SCRIPT_BASE_URL="${SCRIPT_BASE_URL:-https://raw.githubusercontent.com/wso2/agent-manager/main/deployments/scripts}"
```
jhivandb: **Architecture: per-environment Thunder provisioning lives in an ~845-line script fetched from `main` at runtime.**

`SCRIPT_BASE_URL` defaults to `raw.githubusercontent.com/wso2/agent-manager/main/...` and the script `curl`s `add-environment-thunder.sh` at execution time. This has several consequences:

- **Split-brain:** only this script path provisions Thunder. Environments created via the API/console (`CreateEnvironment` has no provisioning step) get no Thunder instance.
- **Version skew:** a deployment on release vN fetches whatever `main` currently serves, which may target a different chart/naming convention than the running service's `naming.go`.
- **Silent fallback:** if the fetch fails (e.g. 404 before this merges to `main`), provisioning falls back to platform Thunder non-fatally.
- Offline/air-gapped runs and supply-chain integrity are also affected.

Provisioning that owns environment lifecycle arguably belongs in the service/operator rather than a remotely-fetched shell script. At minimum, pin the fetched script to the installed release version (as `install-helpers.sh` does) rather than `main`.

deployments/scripts/add-environment.sh
```diff
@@ -172,27 +313,58 @@ RELEASE_NAME="api-platform-${ORG_NAME}-${ENV_NAME}"
...
+        --set "apiGateway.config.policyConfigurations.jwtauth_v1.keymanagers[1].name=ThunderKeyManager"
+        --set "apiGateway.config.policyConfigurations.jwtauth_v1.keymanagers[1].issuer=${THUNDER_ISSUER}"
+        --set "apiGateway.config.policyConfigurations.jwtauth_v1.keymanagers[1].jwks.remote.uri=${THUNDER_INTERNAL_JWKS}"
+        --set "apiGateway.config.policyConfigurations.jwtauth_v1.keymanagers[1].jwks.remote.skipTlsVerify=true"
+    )
+else
+    # Re-assert the platform Thunder keymanager explicitly. Helm --set on an indexed array
+    # replaces the whole list, so keymanagers[0] above already dropped the chart default.
+    HELM_ARGS+=(
+        --set "apiGateway.config.policyConfigurations.jwtauth_v1.keymanagers[1].name=ThunderKeyManager"
```
jhivandb, line 352: **Re-running with `PROVISION_THUNDER=false` can silently invalidate all tokens minted by a live env-Thunder.**

`keymanagers[1]` is a single slot and `helm --set` replaces the whole array entry, so the gateway can't trust both the platform and per-env issuers at once. The existing-release check that sets `THUNDER_PROVISIONED=true` lives only inside the `PROVISION_THUNDER=true` branch — so if an operator re-runs this script (an advertised idempotent workflow) with `PROVISION_THUNDER=false` on an environment that already has a live env-Thunder, this else-branch rewrites `keymanagers[1]` back to the platform issuer/JWKS without checking cluster state. Every JWT previously issued by the still-running env-Thunder then fails validation (401) at the gateway.

Consider gating this rewrite on the actual helm-release state of the env-Thunder, or detecting the live instance before overwriting the slot.

deployments/scripts/add-environment-thunder.sh
```diff
@@ -0,0 +1,845 @@
...
+  echo "  Issuer:          ${issuer}"
+  echo "  JWKS:            ${issuer}/oauth2/jwks"
+  echo "  Trusted issuer:  ${pt_issuer}"
+  echo ""
+  echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
+  echo "  Thunder ID Console — ${ENV_NAME}"
+  echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
+  echo "  URL:      ${issuer}/console"
+  echo "  Username: admin"
+  echo "  Password: ${admin_password}"
```
jhivandb, line 837: **Security: the generated Thunder superadmin password is printed in plaintext to stdout.**

This `echo` (and the corresponding summary output in `setup-openchoreo.sh` / `install.sh`) prints the admin credential directly. These scripts are documented as `curl`-piped, where stdout routinely lands in CI logs and terminal scrollback — capturing the IdP admin password permanently. I couldn't find a pre-existing precedent for printing credentials this way in the repo, so this appears new here.

Suggest printing a `kubectl get secret ... -o jsonpath` retrieval command instead of the secret value itself.

deployments/setup/setup-openchoreo.sh
```diff
@@ -543,3 +588,33 @@ echo ""
...
+  if [ -z "$_host" ]; then
+    # Skip advertising URL details if the routing HTTPRoute is not yet created.
+    continue
+  fi
+  _pass="$(kubectl get secret "$_secret" -n "$_ns" -o jsonpath='{.data.password}' 2>/dev/null | base64 -d)"
```
jhivandb: **This unguarded command substitution can abort the entire setup after everything is already installed.**

The script runs under `set -euo pipefail`, and command-substitution failure in an assignment propagates under `set -e`. If the `password` key is missing/undecodable or `kubectl` has a transient blip here, this line aborts the whole script at the final summary step — after a successful install. Add a `|| true` guard (or `|| _pass="<unavailable>"`).

(Minor, same block: `_active_count` is incremented and the banner header printed before the empty-host `continue`, so the summary can open with an instance it then skips.)

agent-manager-service/clients/thundersvc/naming.go
```diff
@@ -0,0 +1,239 @@
...
+	// system client's OAuth2 secret.
+	thunderSystemClientSecretKey = "client-secret"
+
+	thunderInternalPort = 8090
+	maxReleaseNameLen   = 53
+	truncatePrefixLen   = 46
+)
+
+// ThunderReleaseName returns the Helm release name (and namespace) for an env-Thunder instance.
+// Mirrors thunder_release_name() in add-environment-thunder.sh — must stay in sync.
```
jhivandb: **Maintainability: this naming algorithm is re-implemented in 5 places that must stay byte-identical.**

The same release-name / namespace / host derivation (53-char cap, truncate-to-46 + sha256-6, hyphen trimming) lives here in `naming.go` and in four shell scripts — `add-environment-thunder.sh`, `add-environment.sh`, `remove-environment-thunder.sh`, and `install-helpers.sh` — each carrying its own "must stay in sync" comment.

Because the Go service computes the namespace/service URL from *this* copy while the scripts deploy under *theirs*, any single-copy edit (e.g. changing the length cap or hash width) makes the service address a namespace that was never created — silently, with no error pointing at the cause. Long org/env names near the truncation boundary are exactly the untested edge where drift bites.

Suggest extracting the bash copies into one sourced helper (the scripts already fetch each other), leaving just the bash lib + Go to keep aligned.

yashed: Fixed — extracted the bash side into `deployments/scripts/thunder-naming.sh`, the single source of truth for `thunder_release_name`/`thunder_namespace`/`thunder_host`/`thunder_issuer`.

deployments/setup/setup-openchoreo.sh
yashed: fixed. Added `|| _pass="<unavailable>"` so a transient `kubectl` failure or missing key at this final summary step can't abort the script after a successful install. Also reordered so `_active_count`/the banner header are only touched after the empty-`_host` check, fixing the count/banner-before-skip issue you flagged.

deployments/scripts/add-environment-thunder.sh
yashed: This is requested in architecture design meetings without giving the default _admin:admin_ add a generated password and show it after setup in console

deployments/scripts/add-environment.sh
yashed: Fix: moved the cluster-state check (add-environment.sh:249-262) out of the PROVISION_THUNDER gate so it always runs — only the actual (re-)provisioning attempt stays conditional on that flag.

deployments/scripts/add-environment.sh
yashed: Fixed, the fetch now pins to the release tag (amp/vX.Y.Z) derived from the Agent Manager version already passed via CHART_VERSION, instead of main. No version hardcoded; same convention as install-helpers.sh and the console's getScriptRef(). Falls back to main only when unset/dev, matching console behavior.

agent-manager-service/clients/thundersvc/naming.go
yashed: fixed. ThunderHost/ThunderIssuerURL/ThunderExternalTokenURL/ThunderExternalJWKSURL now derive from config instead of hardcoded constants: base domain via THUNDER_HOST_BASE_DOMAIN (default amp.localhost, unchanged for local dev), scheme/port via the existing TLSConfig.EnableTLS flag. VM/production deployments set THUNDER_HOST_BASE_DOMAIN to their own domain and the reported URLs update accordingly

agent-manager-service/clients/thundersvc/naming.go
yashed: Gated the two hardcoded fallbacks (host.docker.internal, 127.0.0.1) behind the existing IS_LOCAL_DEV_ENV flag, same pattern as auth.go. In prod, only cluster-internal + public URL are tried — no dev assumptions, no shared-address false positives. Verified inside the actual container: works via dev fallback with the flag on, correctly returns not-found with it off.

agent-manager-service/services/environment_service.go
yashed: Fixed the probe itself for now, it was treating any HTTP 200 as "Thunder's alive" with no body check, so I added real JWKS validation. Also made the 4 fallback checks run in parallel instead of sequentially, so a dead env fails in ~2s instead of ~8s. Tried trimming to just the external URL, but only host.docker.internal actually works from the container locally, so kept all four, just parallelized.

The bigger fix (persisting provisioning state) is still valid and not done here that needs a DB table + a callback from the provisioning script similar to what we are going to do in agnetID provisioning. Will address this on that PR

deployments/vm/install-advanced.sh
```diff
@@ -235,6 +235,38 @@ run_advanced_install() {
...
+      # doesn't need a custom CA bundle mounted.
+      export SKIP_CA_BUNDLE_TRUST=true
+      ;;
+    selfsigned)
+      # generate_selfsigned_ca wrote its own root CA alongside the leaf cert — env-
+      # Thunder needs THIS specific CA (not local dev's cert-manager one, which
+      # doesn't exist on a VM) to trust platform Thunder's self-signed cert.
+      export PLATFORM_THUNDER_CA_PEM
+      PLATFORM_THUNDER_CA_PEM="$(cat "$(dirname "$TLS_CERT_FILE")/ca.crt")"
+      ;;
```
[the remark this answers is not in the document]
yashed: fixed

jhivandb (approved):

PR #1251, opened 2026-07-05
Add AgentID provisioning with regenerate and revoke endpoints

yashed:
## Purpose
Agent Manager already lets agents connect to resources such as MCP proxies and LLM providers, but agents don't yet have an identity of their own to authenticate with — access today relies on shared or borrowed credentials rather than one scoped to that specific agent. This PR adds that per-agent identity, which future work will use to scope exactly what each agent can call on MCP proxies and other platform resources.

## Goals
- Give every agent (Internal and External) its own OAuth2 `client_credentials` identity, automatically, per environment.
- Add endpoints to view, regenerate, revoke, and explicitly provision that identity.
- Make provisioning resilient — never block agent creation on a Thunder outage, retry automatically.

## Approach
Each agent gets a Thunder OAuth client per environment, provisioned asynchronously right after agent creation and retried automatically until it succeeds. External agents' secrets are shown once and then destroyed; Internal agents can fetch theirs repeatably via a dedicated endpoint. Regenerate rotates and returns a fresh secret for both agent types; revoke is a kill switch with no secret returned.

Full architecture and design decisions: [AgentID Architecture & Design](https://docs.google.com/document/d/1XcD-UdFIOMFDmeAKDaH-9RjuzmL0sM6FEmo5ci3gN9U/edit?usp=sharing)
Integration guide (endpoints, request/response shapes, one-time-claim behavior): [AgentID Integration Guide](https://docs.google.com/document/d/1XcD-UdFIOMFDmeAKDaH-9RjuzmL0sM6FEmo5ci3gN9U/edit?tab=t.vuqj4zs4v64g)

## User stories
- As an operator, I want my agent to have its own credential to authenticate to platform APIs.
- As an operator, I want to retrieve an External agent's secret safely (once) and regenerate it if needed.
- As an operator, I want to revoke a leaked credential immediately.
- As an admin, I want an Internal agent promoted to a new environment to get its identity automatically.

## Release note
Added AgentID: every agent now gets its own OAuth2 identity per environment, with endpoints to view, regenerate, revoke, and provision it.

## Documentation
- N/A

## Training
N/A 

## Certification
N/A 

## Marketing
N/A 

## Automation tests
- Unit tests
  > Full test suite for provisioning, repository, and reconciler logic (success/failure/retry/claim paths). All passing with `-race`.
- Integration tests
  > Full suite passing against real Postgres. Also manually verified live against a running stack across 4 end-to-end scenarios (external/internal agents, promotion, multi-env) — see [scenario test report](tasks/agentid/docs/agentid-scenario-test-report.md).

## Security checks
- Followed secure coding standards? **yes**
- Ran FindSecurityBugs plugin and verified report? **N/A** — Go codebase
- Confirmed no keys/passwords/tokens/secrets committed? **yes**

## Samples
N/A

## Related PRs
*https://github.com/wso2/agent-manager/pull/1214*

## Migrations
Two additive migrations: creates the `agent_thunder_clients` table and adds an audit column. No manual steps — runs automatically on startup.

## Test environment
macOS, Docker/Colima, PostgreSQL 16, Go 1.25, local k3d + docker-compose.

## Learning
N/A

## Summary of changes

* **New Features**
  * Added per-environment agent identity APIs: view, claim one-time secrets, regenerate, revoke, provision, and fetch credentials.
  * Introduced environment-scoped Thunder client resolution with cached provisioning and automatic background retries.
* **Bug Fixes**
  * Improved idempotency and consistent “not found/not available” behavior for identity and credential states.
  * Enhanced input validation and safer shutdown with proper background service stop.
* **Tests / Chores**
  * Added extensive unit/integration coverage, updated test wiring, mocks, and CI environment setup.
  * Added database migrations for agent Thunder bindings and tracking who requested provisioning.


agent-manager-service/api/agent_routes.go
```diff
@@ -47,4 +47,9 @@ func registerAgentRoutes(rr *middleware.RouteRegistrar, ctrl controllers.AgentCo
...
 	rr.HandleFuncWithValidationAndAuthz("POST /orgs/{orgName}/projects/{projName}/agents/{agentName}/publish-kind", rbac.AgentKindCreate, ctrl.PublishKind)
+	rr.HandleFuncWithValidationAndAuthz("GET /orgs/{orgName}/projects/{projName}/agents/{agentName}/identity", rbac.AgentRead, ctrl.GetAgentIdentity)
+	rr.HandleFuncWithValidationAndAuthz("POST /orgs/{orgName}/projects/{projName}/agents/{agentName}/environments/{envID}/identity/regenerate", rbac.AgentUpdate, ctrl.RegenerateAgentIdentitySecret)
+	rr.HandleFuncWithValidationAndAuthz("POST /orgs/{orgName}/projects/{projName}/agents/{agentName}/environments/{envID}/identity/revoke", rbac.AgentUpdate, ctrl.RevokeAgentIdentitySecret)
+	rr.HandleFuncWithValidationAndAuthz("POST /orgs/{orgName}/projects/{projName}/agents/{agentName}/environments/{envID}/identity/provision", rbac.AgentUpdate, ctrl.ProvisionAgentIdentity)
+	rr.HandleFuncWithValidationAndAuthz("GET /orgs/{orgName}/projects/{projName}/agents/{agentName}/environments/{envID}/identity/credentials", rbac.AgentRead, ctrl.GetAgentCredentials)
```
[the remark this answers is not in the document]
yashed: Addressed: changed both routes from rbac.AgentRead to rbac.AgentUpdate since they return secret-bearing data.

agent-manager-service/clients/thundersvc/agent_secret_store.go
```diff
@@ -0,0 +1,152 @@
...
+// all) — so every segment is checked here rather than assuming upstream
+// safety, mirroring secretmanagersvc's own sanitizeSegment guard.
+func (s *agentSecretStore) key(orgName, projectName, envName, agentName string) (string, error) {
+	for _, seg := range []string{orgName, projectName, envName, agentName} {
+		if strings.Contains(seg, "/") {
+			return "", fmt.Errorf("agent secret path segment %q contains invalid character '/'", seg)
+		}
+	}
+	return path.Join("agent-thunder-clients", orgName, projectName, envName, agentName), nil
+}
```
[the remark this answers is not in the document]
yashed: Addressed: extended the segment check to also reject "", ".", and "..", not just /.

agent-manager-service/clients/thundersvc/env_resolver.go
```diff
@@ -0,0 +1,146 @@
...
+		return nil, fmt.Errorf("%w: %s/%s", ErrThunderUnreachable, orgName, envName)
+	}
+	client := newThunderClientWithDialOverride(baseURL, thunderSystemClientID, clientSecret, resolveToHost)
+
+	r.mu.Lock()
+	r.cache[cacheKey] = client
+	r.mu.Unlock()
+
+	return client, nil
+}
```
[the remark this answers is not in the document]
yashed: wrapped the cold-cache path in singleflight.Group so concurrent first-time resolves for the same key share one read.

agent-manager-service/clients/thundersvc/env_resolver.go
```diff
@@ -0,0 +1,146 @@
...
+	cacheKey := orgName + "/" + envName
+
+	r.mu.RLock()
+	if client, ok := r.cache[cacheKey]; ok {
+		r.mu.RUnlock()
+		return client, nil
+	}
+	r.mu.RUnlock()
+
+	secretPath := path.Join(r.openBaoPath, "data", "thunder-system-clients", orgName, envName)
```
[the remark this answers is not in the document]
yashed: Addressed: applied the same segment validation used in agent_secret_store.go, which this path had none of.

agent-manager-service/controllers/agent_controller.go
```diff
@@ -1006,3 +1012,166 @@ func (c *agentController) PublishKind(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		log.Error("GetAgentIdentity: failed to get agent identity", "orgName", orgName, "agentName", agentName, "error", err)
+		handleCommonErrors(w, err, "Failed to get agent identity")
+		return
+	}
+
+	if views == nil {
+		views = []models.AgentIdentityEnvironmentView{}
+	}
+	utils.WriteSuccessResponse(w, http.StatusOK, views)
```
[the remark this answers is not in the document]
yashed: Addressed: split the claim into a new POST .../identity/claim endpoint — GET is now a safe, retryable read reporting hasUnclaimedSecret.

agent-manager-service/repositories/agent_thunder_client_repository.go
```diff
@@ -0,0 +1,234 @@
...
+	// ClearClaim resets claimed_at to NULL — used by regenerate immediately
+	// after storing a brand-new secret, so that new secret is eligible for the
+	// one-time claim again (MarkClaimed from a previous secret must not carry
+	// over and make a *different*, never-yet-seen secret look already-claimed).
+	ClearClaim(id uuid.UUID) error
+
+	// DeleteByAgent removes every binding for the given agent (used when the agent
+	// itself is deleted).
+	DeleteByAgent(orgName, projectName, agentName string) error
+}
```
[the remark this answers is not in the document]
yashed: threaded ctx through all 9 methods, the GORM impl, the regenerated mock, and every call site.

agent-manager-service/repositories/agent_thunder_client_repository.go
```diff
@@ -0,0 +1,234 @@
...
+	// blank, orphaning it. On conflict, GORM leaves c's fields (including ID)
+	// unpopulated from the DB; callers must treat that as "someone else owns
+	// this binding" rather than assume c now reflects a real row.
+	return r.db.Select("*").Clauses(clause.OnConflict{
+		Columns: []clause.Column{
+			{Name: "org_name"}, {Name: "project_name"}, {Name: "agent_name"}, {Name: "environment_name"},
+		},
+		DoNothing: true,
+	}).Create(c).Error
+}
```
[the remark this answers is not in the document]
yashed: Verified not an issue: confirmed live against Postgres that GORM already omits a zero-value primary key from the INSERT and defers to the column's gen_random_uuid() default.

agent-manager-service/repositories/agent_thunder_client_repository.go
```diff
@@ -0,0 +1,234 @@
...
+	err := r.db.Where("org_name = ? AND project_name = ? AND agent_name = ? AND environment_name = ?",
+		orgName, projectName, agentName, environmentName).First(&c).Error
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return nil, ErrAgentThunderClientNotFound
+		}
+		return nil, err
+	}
+	return &c, nil
+}
```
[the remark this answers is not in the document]
yashed: wrapped every unexpected error with fmt.Errorf("<operation>: %w", err) across all 9 methods, leaving the existing not-found sentinel mapping untouched.

agent-manager-service/services/agent_manager.go
```diff
@@ -2815,6 +2853,172 @@ func findLowestEnvironment(promotionPaths []models.PromotionPath) string {
...
+	}
+	visible := allPipelineEnvironmentNames(pipeline.PromotionPaths)
+
+	filtered := make([]models.AgentIdentityEnvironmentView, 0, len(views))
+	for _, v := range views {
+		if visible[v.EnvironmentName] {
+			filtered = append(filtered, v)
+		}
+	}
+	return filtered, nil
```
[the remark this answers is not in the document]
yashed: this described the exact GET-claims-a-secret bug already fixed in #5 — GetIdentityViews no longer claims anything, so there's nothing left to filter before.

agent-manager-service/services/agent_manager.go
```diff
@@ -2815,6 +2853,172 @@ func findLowestEnvironment(promotionPaths []models.PromotionPath) string {
...
+		return nil, err
+	}
+
+	return &models.AgentRegenerateSecretResponse{
+		EnvironmentName:  environmentName,
+		ProvisioningType: ownership,
+		ClientID:         clientID,
+		ClientSecret:     newSecret,
+		Status:           models.AgentRegenerateSecretStatus,
+	}, nil
```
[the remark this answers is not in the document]
yashed: Verified not an issue: environments are an org-wide resource by design, and RBAC plus the existing (org, project, agent, env) DB query already fully scope every mutation to the caller's own agent.

agent-manager-service/services/agent_thunder_provisioning_service.go
```diff
@@ -0,0 +1,502 @@
...
+			// secret) either way, and cleanup can be retried by revoking again.
+			s.logger.Warn("Failed to delete stored secret during revoke", "bindingID", binding.ID, "error", err)
+		}
+	}
+
+	if err := s.repo.UpdateSecretRef(binding.ID, ""); err != nil {
+		return "", err
+	}
+	return binding.ThunderClientID, nil
+}
```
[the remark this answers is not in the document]
yashed: Addressed: added an in-process per-binding mutex around both methods, with a concurrency test proving the fix (and proving the bug without it).

agent-manager-service/services/agent_thunder_reconciler.go
```diff
@@ -0,0 +1,137 @@
...
+		s.logger.Info("Retrying due agent thunder bindings", "count", len(due))
+	}
+	for _, binding := range due {
+		s.provisioning.AttemptProvision(ctx, binding)
+	}
+
+	if err := tx.Commit().Error; err != nil {
+		s.logger.Error("Failed to commit agent thunder reconciler advisory lock transaction", "error", err)
+	}
+}
```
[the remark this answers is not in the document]
yashed: Fixed: the lock now releases right after the FindDue scan, before the loop of Thunder/OpenBao calls, since ClaimForAttempt's per-binding CAS already prevents double-processing.

agent-manager-service/wiring/wire.go
```diff
@@ -459,6 +466,24 @@ func ProvideAIApplicationRepository(db *gorm.DB) repositories.AIApplicationRepos
...
+	return thundersvc.NewEnvThunderResolver(cfg.OpenBao.URL, cfg.OpenBao.Token, cfg.OpenBao.Path)
+}
+
+// ProvideAgentSecretStore creates the raw OpenBao-backed store for AgentID
+// client credentials (see agent_secret_store.go for why this bypasses the
+// SecretManagementClient/SecretReference machinery).
+func ProvideAgentSecretStore(cfg config.Config) (thundersvc.AgentSecretStore, error) {
+	return thundersvc.NewAgentSecretStore(cfg.OpenBao.URL, cfg.OpenBao.Token, cfg.OpenBao.Path)
+}
+
```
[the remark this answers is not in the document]
yashed: Addressed: added a validateOpenBaoConfig check to both constructors so a missing OPENBAO_* value fails fast instead of only failing on first secret read/write.

agent-manager-service/services/agent_thunder_provisioning_service.go
```diff
@@ -0,0 +1,560 @@
...
+				// pass them through so a failure here doesn't lose them (see
+				// the identical reasoning on the secretStore.Store failure below).
+				s.recordFailure(ctx, binding, thunderAgentID, clientID, fmt.Errorf("recover secret for existing agent identity: %w", err))
+				return
+			}
+		}
+	}
+
+	secretRefPath := binding.SecretRefPath
+	if clientSecret != "" {
```
jhivandb, line 342: **Data drift (env‑Thunder ↔ AMS): a partially‑provisioned binding is marked `completed` without ever storing a secret.**

If a first attempt gets past `CreateAgentIdentity` (Thunder now holds the identity + secret) but `secretStore.Store` at line 297 fails, `recordFailure` persists `thunder_agent_id`/`thunder_client_id` and leaves `status = pending`. On the **retry**, `thunderAgentID != ""`, so the entire `if thunderAgentID == "" { … }` block — including the Section 6.8 *regenerate‑to‑recover* path — is skipped and `clientSecret` stays `""`. This `if clientSecret != ""` guard is then false, `Store` is never called, and `UpdateAfterAttempt` marks the binding **`completed` with `secret_ref_path = ""`**.

Result: AMS reports the AgentID as fully provisioned, but no credential was ever stored (`GetAgentCredentials`/`ClaimSecret` return `ErrAgentCredentialNotAvailable` until a manual regenerate), while Thunder holds an orphaned secret AMS never captured — exactly the env‑Thunder ↔ AMS drift this feature is meant to prevent.

Note that `TestAttemptProvision_AlreadyHasThunderAgentID_SkipsCreate` encodes this behavior: it asserts `status == completed` with a store mock that has **no** `StoreFunc` (so `Store` is never called) even though the binding has no `SecretRefPath`.

**Fix:** on the retry path, when `thunderAgentID != "" && binding.SecretRefPath == ""`, regenerate a secret (`RegenerateAgentSecret`) and store it before marking the binding complete — the same recovery already used for the 409 fallback — instead of completing with an empty secret ref.

agent-manager-service/clients/thundersvc/naming.go
```diff
@@ -195,83 +195,133 @@ func isValidJWKS(body []byte) bool {
...
+// cluster-internal candidate always succeeds when Thunder is actually reachable, and
+// probing 127.0.0.1 there would hit agent-manager-service's own pod rather than
+// Thunder, which is both pointless and a latent false-positive risk if anything ever
+// answers on that port. This is the single source of truth for the candidate cascade —
+// both ThunderProbe and ResolveThunderBaseURL (used by EnvThunderResolver to reach
+// env-Thunder's admin API for per-agent client provisioning) build their attempts from
+// this same list, so the two can never drift out of sync with each other.
+func thunderBaseURLCandidates(org, env string) []thunderURLCandidate {
+	host := ThunderHost(org, env)
+	externalBaseURL := fmt.Sprintf("http://%s:8080", host)
```
jhivandb: **Misconfiguration between env‑Thunder's advertised endpoint and what AMS/the gateway use under TLS.**

`externalBaseURL` is hardcoded to `http://<host>:8080` regardless of `config.TLSConfig.EnableTLS`. Everywhere else the external origin comes from `thunderExternalOrigin()`, which returns `https://<host>` (no `:8080`) when TLS is enabled — and that origin is what `ThunderIssuerURL`, `ThunderExternalTokenURL`, and `ThunderExternalJWKSURL` (the JWKS the gateway's ThunderKeyManager fetches, and the `iss` env‑Thunder stamps into agent tokens) all produce.

This is a regression from the pre‑refactor code, which used `ThunderExternalJWKSURL(org, env)` for candidate 2 and therefore honored TLS. Now, on a TLS/VM deployment, if the cluster‑internal DNS candidate is unreachable, both `ResolveThunderBaseURL` (used by `EnvThunderResolver` to reach env‑Thunder's admin API for per‑agent provisioning) and `ThunderProbe` fall back to `http://<host>:8080` — the wrong scheme, and a port the Caddy front‑end doesn't expose — so provisioning fails and a live env‑Thunder is reported unreachable. The base URL AMS talks to also no longer matches env‑Thunder's own advertised origin / the gateway's issuer.

**Fix:** derive the external candidate from `thunderExternalOrigin(org, env)` (which already returns the TLS‑aware scheme/port) rather than a hardcoded `http://…:8080`, so the probe/resolve URL matches the issuer/JWKS the rest of the system relies on. The local‑dev `resolveToHost` fallbacks can keep their fixed `:8080` host since they only apply when `IsLocalDevEnv` (TLS off).

agent-manager-service/clients/thundersvc/naming.go
```diff
@@ -282,16 +332,24 @@ func ThunderProbe(ctx context.Context, org, env string) bool {
...
+	env = slugify(env)
 	project = slugify(project)
 	agent = slugify(agent)
-	if org == "" || project == "" || agent == "" {
-		panic("org, project, and agent names must be valid alphanumeric slugs and not empty")
+	if org == "" || env == "" || project == "" || agent == "" {
+		panic("org, env, project, and agent names must be valid alphanumeric slugs and not empty")
 	}
-	name := fmt.Sprintf("amp-agent-%s-%s-%s", org, project, agent)
+	name := fmt.Sprintf("amp-agent-%s-%s-%s-%s", org, env, project, agent)
```
jhivandb, line 358: **Truncation with no disambiguating hash lets two agents collide onto one Thunder identity (cross‑agent drift/leak).**

Unlike `ThunderReleaseName` and `ThunderHost` — which append a `thunderSHA6` hash when they truncate — `AgentThunderAppName` truncates to `maxAgentAppNameLen` (100) with a plain `strings.TrimRight(name[:maxAgentAppNameLen], "-")` and no collision‑avoidance suffix. Adding `env` to the name in this PR makes the string longer, so truncation (and thus collisions) become more likely.

Two agents whose `amp-agent-<org>-<env>-<project>-<agent>` slugs share the first 100 characters produce the **same** Thunder app name. When the second provisions, `CreateAgentIdentity` gets `409 AGT-1013`, `findAgentByName` returns the *first* agent's Thunder ID/clientID (exact‑name match), and the second binding is stored pointing at the first agent's identity. Both AMS bindings now reference one Thunder AgentID — regenerating or revoking one rotates/kills the other's credential.

**Fix:** append a `thunderSHA6(org+"/"+env+"/"+project+"/"+agent)` suffix on truncation, matching the collision‑safe scheme already used by `ThunderReleaseName`/`ThunderHost`.

yashed: fixed - truncation now appends a thunderSHA6(org/env/project/agent) suffix on overflow, matching the collision-safe scheme already used by ThunderReleaseName/ThunderHost

agent-manager-service/clients/thundersvc/naming.go
yashed: fixed - the external-ingress candidate now uses thunderExternalOrigin(org, env) instead of a hardcoded http://<host>:8080,

agent-manager-service/services/agent_thunder_provisioning_service.go
yashed: added a retry path (thunderAgentID != "" && SecretRefPath == "") that calls RegenerateAgentSecret and stores it before completing, reusing the same recovery logic already used for the 409 conflict fallback;

agent-manager-service/services/agent_manager_test.go
```diff
@@ -142,3 +169,107 @@ func TestValidateEffectivePair_NoPythonIsNoOp(t *testing.T) {
...
+			t.Fatal("must not attempt to claim a secret when the agent itself cannot be fetched")
+			return "", "", "", nil
+		},
+	}
+	s := &agentManagerService{ocClient: ocClient, agentThunderProvisioning: stub, logger: slog.Default()}
+
+	_, err := s.ClaimAgentIdentitySecret(context.Background(), "acme", "proj1", "my-agent", "dev")
+
+	require.Error(t, err)
+}
```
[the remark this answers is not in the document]
yashed: fixed

jhivandb (commented):
Code review of the agent-creation identity-sync mechanism (write-ahead + background provisioning + reconciler-retry, plus the Thunder/OpenBao clients and repository). 7 correctness/robustness findings and 3 cleanups, inline below.

agent-manager-service/services/agent_thunder_provisioning_service.go
```diff
@@ -0,0 +1,568 @@
...
+// phase recognizes (Section 6.4) — everything else is retried up to
+// maxProvisionAttempts times before being marked FAILED. resolvedAgentID/
+// resolvedClientID carry through a Thunder identity that was already created
+// before a LATER step in this same attempt failed (e.g. secret storage) —
+// passing "" for either is correct when nothing was resolved yet.
+// UpdateAfterAttempt only writes a non-empty field, so this never clobbers an
+// already-stored value with a blank one.
+func (s *agentThunderProvisioningService) recordFailure(ctx context.Context, binding models.AgentThunderClient, resolvedAgentID, resolvedClientID string, cause error) {
+	attemptsSoFar := binding.AttemptCount + 1
+	permanent := errors.Is(cause, thundersvc.ErrThunderNotProvisioned)
```
jhivandb: **Issue — `ErrThunderNotProvisioned` is terminal; an identity gap never heals.** It's classified as permanent here, so the binding is marked `FAILED` on the very first attempt. Nothing recovers it: `FindDue` (`agent_thunder_client_repository.go:164-178`) only selects `PENDING` and stale `IN_PROGRESS` rows, never `FAILED`; and `ProvisionForEnvironmentIfMissing` (`:220-223`) treats *any* existing row — including `FAILED` — as `alreadyExisted=true` and returns without retrying. So the common window where an env exists before its (async) env-Thunder bootstrap has finished produces a permanent, silent identity gap for that env, directly contradicting the interface promise at `:79-80` that "the reconciler already owns retrying anything not yet completed."

**Fix:** Don't treat `ErrThunderNotProvisioned` as terminal — it means "not bootstrapped *yet*." Either drop it from the `permanent` classification so it retries like any transient cause, or give it its own long-backoff retry lane; and let `ProvisionForEnvironmentIfMissing` / the re-provision endpoint re-open a `FAILED` row (reset to `PENDING`) so an explicit call can heal it once bootstrap completes.

agent-manager-service/services/agent_thunder_provisioning_service.go
```diff
@@ -0,0 +1,568 @@
...
+		b := models.AgentThunderClient{
+			OrgName:          orgName,
+			ProjectName:      projectName,
+			AgentName:        agentName,
+			EnvironmentName:  env,
+			ProvisioningType: ownership,
+			Status:           models.AgentThunderStatusPending,
+			RequestedBy:      requestedBy,
+		}
+		if err := s.repo.Upsert(ctx, &b); err != nil {
```
jhivandb: **Issue — a transient DB error during write-ahead leaves an env with no row and no retry path.** If `repo.Upsert` fails for one env, the code logs and `continue`s, so that env is never appended to `bindings` and never attempted. The whole write-ahead durability story rests on the row existing for the reconciler to find — a momentary DB blip on that one `Upsert` means the identity for that env is silently never provisioned and never retried, and the caller can't tell because `ProvisionForAgent` returns nothing.

**Fix:** Make the write-ahead durable at the write: either fail the whole operation (return an error) so the caller can retry, or record the failed env somewhere the reconciler can pick it up. At minimum, don't silently swallow — the guarantee has a hole exactly at the point it's supposed to hold.

agent-manager-service/services/agent_thunder_provisioning_service.go
```diff
@@ -0,0 +1,568 @@
...
+			s.logger.Warn("Failed to delete Thunder agent identity", "agentName", agentName, "env", b.EnvironmentName, "error", err)
+		}
+		if b.SecretRefPath != "" {
+			if err := s.secretStore.Delete(ctx, b.SecretRefPath); err != nil {
+				s.logger.Warn("Failed to delete stored agent secret", "agentName", agentName, "env", b.EnvironmentName, "error", err)
+			}
+		}
+	}
+
+	if err := s.repo.DeleteByAgent(ctx, orgName, projectName, agentName); err != nil {
```
jhivandb: **Issue — background `DeleteAllBindings` races delete-then-recreate of a same-named agent and wipes the new bindings.** `DeleteAgent` fires this as `go DeleteAllBindings(context.WithoutCancel(ctx), …)` (`agent_manager.go:2078`/`2091`). The goroutine snapshots bindings via `FindByAgent`, slowly deletes each Thunder identity, then calls `DeleteByAgent(org, project, agent)` — which is unscoped by ID/time (`agent_thunder_client_repository.go:249`) and deletes *all* rows for that agent name. If a same-named agent is recreated while this runs, that final delete wipes the recreated agent's fresh binding rows too, leaving orphaned Thunder identities and an agent with no bindings.

**Fix:** Scope the final delete to what was actually snapshotted — delete by the specific binding IDs (`id IN (…)`) captured in `FindByAgent`, rather than a blanket delete-by-agent-name, so a concurrent recreation isn't collateral.

agent-manager-service/clients/thundersvc/env_resolver.go
```diff
@@ -0,0 +1,174 @@
...
+			return nil, ErrThunderNotProvisioned
+		}
+
+		baseURL, resolveToHost, ok := r.resolveBaseURL(ctx, orgName, envName)
+		if !ok {
+			return nil, fmt.Errorf("%w: %s/%s", ErrThunderUnreachable, orgName, envName)
+		}
+		client := newThunderClientWithDialOverride(baseURL, thunderSystemClientID, clientSecret, resolveToHost)
+
+		r.mu.Lock()
```
jhivandb, line 173: **Issue — the resolver caches the Thunder client forever; a rotated env-Thunder system secret breaks all provisioning for that env until restart.** One `ThunderClient` is cached per org/env, built from the system-client secret read once from OpenBao, and never invalidated (no TTL, no error-triggered eviction). If that env-Thunder's system-client secret is rotated (re-bootstrap), every subsequent provision/regenerate/revoke for that env authenticates with the stale secret and fails until the AMS process restarts.

**Fix:** Add invalidation — a TTL on the cache entry, and/or evict on an auth failure (401 / `invalid_client`) from the underlying client so the next call re-reads OpenBao and rebuilds with the current secret.

agent-manager-service/services/agent_thunder_reconciler.go
```diff
@@ -0,0 +1,144 @@
...
+	// not the slow Thunder/OpenBao calls below.
+	if err := tx.Commit().Error; err != nil {
+		s.logger.Error("Failed to commit agent thunder reconciler advisory lock transaction", "error", err)
+		return
+	}
+
+	if len(due) > 0 {
+		s.logger.Info("Retrying due agent thunder bindings", "count", len(due))
+	}
+	for _, binding := range due {
```
jhivandb, line 151: **Issue — the reconciler processes its 50-row batch strictly sequentially, so one unreachable env starves the whole sweep.** Each `AttemptProvision` can block on `Resolve` probing candidates one-by-one (`naming.go:287`, up to 4 × 2s for an unreachable env) plus Thunder HTTP calls (30s timeout each). A batch containing bindings for one slow/unreachable env can run for minutes; the 1-minute ticker simply drops ticks while busy, pushing healthy bindings' retries well past the 15-minute SLA the backoff comment advertises.

**Fix:** Run the per-binding attempts with bounded concurrency (a small worker pool / `errgroup` with a limit). `ClaimForAttempt` already makes each attempt independent and safe to run in parallel, so one unreachable env can't starve the rest of the sweep.

agent-manager-service/services/agent_thunder_provisioning_service.go
```diff
@@ -0,0 +1,568 @@
...
+		bindings = append(bindings, b)
+	}
+
+	// Detach from the request context so the background attempt survives the
+	// HTTP handler returning, mirroring the existing deleteAgentLLMConfigurations
+	// pattern in agent_manager.go.
+	go s.attemptAll(context.WithoutCancel(ctx), bindings)
+}
+
+func (s *agentThunderProvisioningService) attemptAll(ctx context.Context, bindings []models.AgentThunderClient) {
```
jhivandb, line 238: **Issue — no panic recovery in the detached provisioning goroutine or the reconciler loop; one panicking attempt crashes the whole service.** Neither this `attemptAll` (go-launched, detached) nor the reconciler's `for … AttemptProvision` loop (`agent_thunder_reconciler.go:83`/`141`) has a `defer recover()`. Any panic inside an attempt — e.g. `AgentThunderAppName`'s `panic(...)` at `naming.go:356` on a name that slugifies to empty, or any future nil-map/deref — takes down the entire AMS process. For the reconciler, the offending `IN_PROGRESS` binding is re-selected once it goes stale, so it crash-loops. (The `AgentThunderAppName` trigger itself is low-reachability since org/env are normally valid slugs; the real exposure is that a background credential job should never be able to crash the API server.)

**Fix:** Wrap each independent attempt in a `defer recover()` that logs and marks just that one binding failed/retryable, so a single bad binding is isolated instead of killing the process (and the loop) for every agent.

agent-manager-service/services/agent_thunder_provisioning_service.go
```diff
@@ -0,0 +1,568 @@
...
+}
+
+// keyedMutex serializes RegenerateSecret/RevokeSecret per binding within this
+// process, so two concurrent rotations for the same binding can't interleave
+// their Thunder call and OpenBao write and leave the stored secret mismatched
+// with what Thunder actually considers active. In-process only — it does not
+// protect across multiple replicas of this service.
+type keyedMutex struct {
+	mu    sync.Mutex
+	locks map[string]*sync.Mutex
```
jhivandb: **Issue — `keyedMutex.locks` grows without bound.** Every distinct binding that goes through `RegenerateSecret`/`RevokeSecret` adds a permanent entry to the `locks` map; nothing evicts them. Over a long-lived process with many agents/environments this is a slow memory leak.

**Fix:** Evict the entry when the last waiter releases (refcount the mutex and delete at zero), or use a `sync.Map` with periodic cleanup, so the map doesn't grow unbounded with the number of bindings ever rotated.

agent-manager-service/clients/thundersvc/agent_secret_store.go
```diff
@@ -0,0 +1,173 @@
...
+	return nil
+}
+
+// NewAgentSecretStore creates an AgentSecretStore backed by a real OpenBao
+// server at openBaoURL, authenticating with openBaoToken.
+func NewAgentSecretStore(openBaoURL, openBaoToken, openBaoPath string) (AgentSecretStore, error) {
+	if err := validateOpenBaoConfig(openBaoURL, openBaoToken, openBaoPath); err != nil {
+		return nil, err
+	}
+	cfg := vault.DefaultConfig()
```
jhivandb: **Issue — OpenBao client bootstrap and KV-v2 path/unwrap logic is duplicated (cleanup).** This `NewAgentSecretStore` and `NewEnvThunderResolver` (`env_resolver.go:88`) duplicate the same OpenBao client-construction sequence (`vault.DefaultConfig` → `NewClient` → `SetToken` → `Logical()`), and both re-implement the KV-v2 `data`/`metadata` path convention and the `secret.Data["data"].(map[string]any)` unwrap that `secretmanagersvc`'s OpenBao provider already encapsulates. Their inline path-segment validators are also weaker than the existing `sanitizeSegment` (they check only `""`/`.`/`..`/`/`, not the charset). (`validateOpenBaoConfig` itself is already shared across the two — only the client construction and KV plumbing are duplicated.)

**Fix:** Extract one `newOpenBaoLogical(url, token)` helper plus shared raw-KV read/write/delete helpers (and reuse `sanitizeSegment`), so a future KV/auth change lands in one place instead of three.

agent-manager-service/clients/thundersvc/agent_client.go
```diff
@@ -0,0 +1,292 @@
...
+	return "", "", fmt.Errorf("thunder list agents exceeded %d pages looking for %s", maxPages, name)
+}
+
+// RegenerateAgentSecret generates a new client secret and applies it to the
+// existing AgentID. Thunder's /agents API has no dedicated regenerate endpoint —
+// PUT only auto-regenerates a secret when transitioning from a non-secret auth
+// method, which does not apply to an already-secret-based client_credentials
+// agent — so this explicitly supplies the new value, exactly like
+// RegenerateClientSecret does for /applications.
+func (c *thunderClient) RegenerateAgentSecret(ctx context.Context, thunderAgentID string) (string, error) {
```
jhivandb, line 132: **Issue — `agent_client.go` re-implements the existing `/applications` Thunder helpers (cleanup).** `RegenerateAgentSecret` is a near line-for-line copy of `regenerateSecret` (`client.go:324`) — GET → map → `setInboundClientSecret` → `delete(id)` → PUT → decode; `findAgentByName` (`:122`) duplicates `findApp`'s list-page pagination (`client.go:441`); and `DeleteAgentIdentity` (`:253`) hand-rolls request/close/status handling that existing helpers already centralize — `doRequest` (`identity_client.go:1071`, request lifecycle + typed 404) and `deleteApp` (`client.go:505`, the idempotent 404→`false` delete).

**Fix:** Parameterize the resource segment (`agents` vs `applications`) on the existing helpers rather than forking them, so this security-sensitive secret-rotation path can't silently diverge from its `/applications` twin.

agent-manager-service/services/agent_thunder_provisioning_service.go
```diff
@@ -0,0 +1,568 @@
...
+// next attempt — a flat 3 minutes for every retry (no escalation to longer
+// delays; a real outage is retried just as promptly on attempt 4 as attempt
+// 1, since env-Thunder recovering doesn't get less likely over time). Four
+// retries at 3 minutes apart means the 5th and final attempt starts at most
+// 12 minutes after the first, so the whole retry budget for one binding
+// resolves — success or FAILED — within the 15-minute SLA. Index 0 is unused
+// (there is no "0 attempts made" retry); the last entry is reused for any
+// attempt count beyond it, though maxProvisionAttempts stops that from
+// actually happening.
+var provisionBackoffSchedule = []time.Duration{
```
jhivandb: **Issue — dead-weight backoff table and an invisible "empty means keep" update contract (cleanup).** `provisionBackoffSchedule` is five identical 3-minute values plus an unused index 0 and an 11-line comment explaining the off-by-one — and the slice must be kept in sync with `maxProvisionAttempts` by hand or `recordFailure`'s `provisionBackoffSchedule[attemptsSoFar]` panics. Separately, `AgentThunderAttemptUpdate` (`agent_thunder_client_repository.go:96`) treating empty `ThunderAgentID`/`ThunderClientID`/`SecretRefPath` as "don't write" is an invisible contract (it already forced a separate `UpdateSecretRef` method because `""` can't mean "clear"). Minor: the `owner` param of `CreateAgentIdentity` is dead (only caller passes `""`, `:271`), and `findAgentByName`'s inner `var page struct` shadows the loop counter `page`.

**Fix:** Replace the slice with a single `const provisionRetryDelay = 3 * time.Minute`; make the update fields `*string` so set-vs-keep is explicit (removing the need for the separate `UpdateSecretRef` special-case); drop the dead `owner` param; and rename the shadowed `page` struct.

agent-manager-service/clients/thundersvc/agent_client.go
yashed: Fix (`client.go`, `agent_client.go`): parameterized the resource segment on the existing helpers rather than forking them

agent-manager-service/clients/thundersvc/agent_secret_store.go
yashed: - Extracted `newOpenBaoLogical(url, token)` and used by both `NewAgentSecretStore` and `NewEnvThunderResolver`.
- Extracted `readOpenBaoKVv2Data(ctx, reader, basePath, keySegments...)` that used by both `agentSecretStore.Get` and `envThunderResolver.Resolve`, returning `(nil, nil)` on a missing secret so each caller maps that to its own existing sentinel (`ErrAgentSecretNotFound` / `ErrThunderNotProvisioned`) unchanged.
- Unified the duplicate `openBaoReader`/`openBaoReadWriter` interfaces — `openBaoReadWriter` now embeds `openBaoReader` instead of redeclaring the same method.

agent-manager-service/services/agent_thunder_provisioning_service.go
```diff
@@ -0,0 +1,623 @@
...
+	// Thunder cleanup can orphan a Thunder identity, an accepted, harmless
+	// tradeoff already made elsewhere in this file.
+	ids := make([]uuid.UUID, 0, len(bindings))
+	for _, b := range bindings {
+		ids = append(ids, b.ID)
+	}
+	if err := s.repo.DeleteByIDs(ctx, ids); err != nil {
+		s.logger.Error("Failed to delete agent thunder client rows", "agentName", agentName, "error", err)
+	}
+
```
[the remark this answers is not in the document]
yashed: fixed

jhivandb (approved):

Issue #1262, opened 2026-07-07
[Task]: Inject AgentID credentials into platform-hosted agent pods


yashed:
### Please select the area the issue is related to

Area/Agent Lifecycle (Agent create, configure, kinds, build, test, deploy, keys/tokens)

### Please select the aspect the issue is related to

Aspect/Identity & Access (Identities, identity providers, roles, permissions, tenancy)

### Description

**Summary**
AgentID provisioning already creates a Thunder OAuth2 identity, meaning a client_id and a client_secret, for every internal agent in every environment. This issue covers the next step in building AgentID: automatically injecting those credentials into the agent's running pod as environment variables. The four values to inject are the client_id, the client_secret, the token endpoint, and the scopes. Minting and using the actual token happens inside the agent's own code at runtime

**Motivation**
An internal agent already has a real Thunder identity, but there is currently no way for that identity to reach the running pod automatically. Without this step, AgentID cannot actually be used at runtime, since an agent has no way to obtain its own credentials. Completing this step lets agents start using their real identity instead of continuing to rely on the shared static keys currently used for LLM and MCP proxy access, which have no real identity and no scopes behind them.

**Scope**

- [ ] - Inject client_id, client_secret (through a secret reference), token_endpoint, and scopes into the pod, reusing the existing LLM and MCP proxy key injection pattern, which is an OpenChoreo SecretReference combined with a ReleaseBinding update.
- [ ] - Wire the injection into the agent create, deploy, and promote flows.
- [ ] - Wire credential refresh into the existing regenerate and revoke endpoints, since today they only rotate the secret in Thunder and never update the running pod.
- [ ] - Confirm that gateway JWKS validation already works per environment, and fix it if it does not.

**Out of scope**
- MCP server to scope mapping and gateway scope authorization
- Minting, caching, and refreshing the token, and attaching it to gateway calls, since this runs inside the agent's own code at runtime.
- A shared token fetch helper library for agent developers, which can be its own follow up issue if needed.

**Related Issues**
- https://github.com/wso2/agent-manager/issues/1253
- https://github.com/wso2/agent-manager/issues/1254

0xbrainkid:
This is a good place to make the runtime identity lifecycle auditable, not just injectable.

Two implementation checks that may prevent drift later:

- include an identity version / secret version in the pod annotation or release binding, so regenerate/revoke can prove which running pods still carry stale AgentID material
[...]

PR #1274, opened 2026-07-08
 Inject AgentID credentials into platform-hosted agent pods

yashed:
## Purpose
AgentID provisioning already creates a Thunder OAuth2 identity (`client_id` + `client_secret`) for every internal (platform-hosted) agent, in every environment. The last step is actually getting that identity into the agent's running pod and it was not implemented, so an internal agent had no way to obtain its own credentials at runtime without a human manually calling the identity API and wiring the result in by hand.

- Related Issue - https://github.com/wso2/agent-manager/issues/1262

## Goals
- Automatically inject an internal agent's `client_id`, `client_secret`, token endpoint, and scopes into its pod as environment variables.
- Keep credentials strictly isolated per environment: promoting an agent must never let one environment's credentials reach another environment's pod.
- Propagate credential rotation and revocation to the running pod, not just to Thunder.
- Reuse the platform's existing, already-proven credential-injection mechanism (the same one used for LLM provider and MCP proxy keys) rather than building a new one.

## Approach
- New `AgentIdentityInjectionService` builds four env vars per agent per environment: client ID, client secret, token endpoint, and scopes.
- Reuses the same OpenChoreo mechanism already used for LLM/MCP proxy keys and no new injection pattern.
- Client secret is never a literal and delivered via a `SecretReference` → Kubernetes Secret → `SecretKeyRef`, pointing at the same OpenBao path the AgentID provisioning service already writes to (no duplicate storage).
- Wired into every place env vars get written, so credentials stay current: deploy, promote, rotation, revoke, config updates, and deletion.
- Promotion is blocked if the target environment's identity isn't ready yet, preventing a pod from inheriting another environment's credentials.
- Rotation forces an immediate secret refresh and pod restart; revoke removes the credential entirely.
- OpenChoreo writes retry on conflict, so a concurrent update can't silently drop a change.
-  **Scopes are now resolved **: the requested scope list is the union of every scope defined on any MCP proxy the agent is bound to in that environment, read directly from AMS's own Postgres. Two best-effort live-refresh hooks keep an already-running pod's scopes current without waiting for its next deploy/promote: one fires when the agent's own MCP proxy bindings change, the other when a proxy it's bound to has its own scopes edited.
- Wired into every place env vars get written, so credentials stay current: deploy, promote, rotation, revoke, config updates, and deletion.

## User stories
- As a platform admin, I want an internal agent's identity credentials delivered into its pod automatically, so nobody has to manually fetch and configure them.
- As a platform admin, I want a credential rotation or revocation to actually take effect on the running agent, not just in Thunder.
- As a platform admin, I want an agent's credentials in one environment to never leak into another environment when it's promoted.

## Release note
N/A

## Documentation
N/A

## Training
N/A

## Certification
N/A

## Marketing
N/A

## Automation tests
 - Unit tests
   > 18 new tests covering env var construction, `SecretReference` handling, all skip states, retry logic, rotation uniqueness, and name collision-avoidance. Additional tests for the provisioning hook (with `-race`), and promotion isolation (blocks when not ready, excludes source/includes target credentials).
 - Integration tests
   > Existing promotion tests updated to seed a ready identity for the new readiness check. Full `make test-unit` and `make test-integration` pass with zero failures.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A

## Samples
N/A

## Related PRs
- https://github.com/wso2/agent-manager/pull/1214
- https://github.com/wso2/agent-manager/pull/1251

## Migrations (if applicable)
N/A

## Test environment
macOS, Docker/Colima, PostgreSQL 16, Go 1.25, local k3d + docker-compose.
 
## Learning
N/A

## Summary

* **New Features**
  * Added automatic per-environment AgentID credential injection for internal agents across deploys, promotions, and configuration/provisioning.
  * Added initial backfill and periodic reconciliation for recently completed internal bindings (keyset pagination).
  * Added `workloadRefreshWarning` to regenerate/revoke responses when workload updates can’t be applied.

* **Bug Fixes**
  * Improved SecretReference handling so rotated credentials re-sync immediately via template annotation changes.
  * Enhanced reconciliation robustness (deterministic behavior, bounded retries) and ensured external cleanup continues despite cleanup failures.
  * Extended proxy-bound MCP scope refresh to reconcile identity for affected environments/agents.


agent-manager-service/services/agent_manager.go
```diff
@@ -3290,6 +3348,65 @@ func (s *agentManagerService) PromoteAgent(ctx context.Context, ouID string, pro
...
+	// org AFTER this agent was created, so its AgentID binding could be
+	// missing entirely (agent-creation only provisions into environments that
+	// existed at that time). Doing this BEFORE building envOverrides below —
+	// rather than after promotion succeeds, as this used to work — matters:
+	// see the hard block right after this call.
+	var requestedBy string
+	if callerClaims := jwtassertion.GetTokenClaims(ctx); callerClaims != nil {
+		requestedBy = callerClaims.Sub
+	}
+	if _, err := s.agentThunderProvisioning.ProvisionForEnvironmentIfMissing(
```
jhivandb: **Critical — panic when AgentID provisioning is not configured.** This is the only call site in this file that dereferences `s.agentThunderProvisioning` without a nil guard (compare lines 1302, 2189, 2204, 3051, 3081, 3117, 3150, 3196, 3251, and the post-promote call at 3591). `app.go` explicitly supports `opts.AgentThunderProvisioning == nil`, in which case this field is a nil interface and **every promotion panics**.

And guarding just this call isn't enough: with provisioning disabled, no binding ever exists, `EnvVarsForEnvironment` returns nil, and the hard block below permanently rejects every promotion. The whole identity block (provision-if-missing + env var build + hard block) needs to be skipped when `s.agentThunderProvisioning == nil`.

Separately: this call makes the pre-existing post-promote `ProvisionForEnvironmentIfMissing` (line ~3596, guarded, after `PromoteComponent`) dead weight — the hard block guarantees the target binding exists and is COMPLETED before promotion runs. One of the two should go.

agent-manager-service/services/agent_manager.go
```diff
@@ -3290,6 +3348,65 @@ func (s *agentManagerService) PromoteAgent(ctx context.Context, ouID string, pro
...
+	// lowest environment are written into the shared OpenChoreo Workload CR
+	// (inherited by every environment unless a ReleaseBinding overrides it).
+	// If the target environment's own identity isn't ready yet, tgtIdentityEnvVars
+	// is empty and NOTHING would override those keys in the target's
+	// ReleaseBinding — the promoted pod would silently start up presenting the
+	// SOURCE (or lowest) environment's real client_id/client_secret, a genuine
+	// cross-environment credential leak. Refusing to promote until the
+	// target's own identity is ready is the same strategy already used a few
+	// lines above for LLM/system configuration (len(tgtSystemKeys) == 0 case).
+	if len(tgtIdentityEnvVars) == 0 {
```
jhivandb, line 3476: **The hard block conflates four distinct states.** `EnvVarsForEnvironment` returns nil for: no binding, provisioning pending/failed, external agent, and *revoked* credential. The error message ("still provisioning … retry once it shows status=completed") is wrong for several of them:

- **Revoked target credential** (`SecretRefPath == ""`): promotion is permanently blocked and the message tells the user to wait for something that will never happen — they need to *regenerate*, not retry.
- **FAILED provisioning**: same, retrying never helps.
- **First promote to a new environment always fails**: `ProvisionForEnvironmentIfMissing` kicks off provisioning on a detached goroutine, so this check virtually always sees PENDING immediately after. Provisioning completes in seconds — consider a short bounded poll before giving up, otherwise a common operation becomes a guaranteed error-then-retry.

At minimum, read the binding status here and produce a state-specific error.

agent-manager-service/services/agent_identity_injection_service.go
```diff
@@ -0,0 +1,612 @@
...
+// templateAnnotations may be nil; see secretRotatedAtAnnotation for when it
+// is not.
+func (s *agentIdentityInjectionService) ensureSecretReference(ctx context.Context, binding *models.AgentThunderClient, templateAnnotations map[string]string) (string, error) {
+	refName := AgentIdentitySecretRefName(binding.AgentName, binding.EnvironmentName)
+	req := client.CreateSecretReferenceRequest{
+		Namespace:           binding.OUID,
+		Name:                refName,
+		ProjectName:         binding.ProjectName,
+		ComponentName:       binding.AgentName,
+		KVPath:              binding.SecretRefPath,
```
jhivandb: **Architecture — this line couples a second service to OpenBao, outside the deployment seam.** `AgentThunderProvisioningService` is deliberately pluggable (injected per-deployment via `app.Options.AgentThunderProvisioning`; the open-source build passes `NewOpenBaoAgentThunderProvisioning`). That factory is supposed to be the only place that knows secrets live in OpenBao.

This service breaks that: it reads `binding.SecretRefPath` raw from the shared repository and feeds it to `CreateSecretReferenceRequest.KVPath` — whose own doc comment says "Path in OpenBao KV store" — and hardcodes the KV key layout via `thundersvc.AgentSecretKeyClientSecret`. It is wired **unconditionally** in `wire.go` and consumed unconditionally by deploy/promote/config-update flows.

Consequence: a deployment plugging in a non-OpenBao provisioning implementation (e.g. AWS Secrets Manager) compiles fine and then silently creates SecretReference CRs pointing at garbage paths at runtime. The two services are coupled *through a DB column's semantics*, which no interface expresses.

Suggestion: have the provisioning service own the "where is this credential" question — e.g. add `CredentialLocation(binding) (kvPath, key string, ok bool)` to `AgentThunderProvisioningService` (or a small locator interface it implements) and have this service ask it instead of interpreting `SecretRefPath` itself. Alternatively, provide this injection service via `app.Options` alongside the provisioning factory so deployments supply a matching pair.

agent-manager-service/services/agent_thunder_provisioning_service.go
```diff
@@ -129,27 +129,77 @@ type AgentThunderProvisioningService interface {
...
 	// they have no claim state, and use GetCredentials instead.
 	ClaimSecret(ctx context.Context, ouID, projectName, agentName, envName string) (agentID, clientID, clientSecret string, err error)
+
+	// SetWorkloadInjector backfills the workloadInjector once the real
+	// AgentIdentityInjectionService exists (this service is constructed before
+	// the OpenChoreo client — see NewOpenBaoAgentThunderProvisioning). app.Run
+	// calls it exactly once, before the reconciler or HTTP server start, so
+	// every later AttemptProvision observes the field already set with no
+	// concurrent read/write. No-op when injector is nil.
+	SetWorkloadInjector(injector AgentIdentityInjectionService)
```
jhivandb: **Interface pollution on the deployment seam.** These two methods exist only to work around construction order in `app.Run`, but every alternative provisioning implementation is now forced to implement a mutable setter and a reconcile hook that may be meaningless for it.

Two cheaper shapes:
- Keep `SetWorkloadInjector` off the core interface: have `app.Run` type-assert an optional interface (`interface{ SetWorkloadInjector(...) }`) — implementations that don't do injection simply don't implement it.
- `ReconcileWorkloadInjection` doesn't need to be on this interface for the reconciler's sweep — the reconciler could hold the injector directly. Only the post-`AttemptProvision` hook genuinely needs it inside this service.

Also worth noting: setter-based two-phase init is a latent race generator. The doc comment argues why it's currently race-free, but that argument depends on call ordering in `app.Run` that nothing enforces.

agent-manager-service/services/agent_configuration_service.go
```diff
@@ -2396,9 +2413,26 @@ func (s *agentConfigurationService) updateMCPConfig(ctx context.Context, existin
...
 	return s.GetMCP(ctx, existingConfig.UUID, ouID, projectName, agentName)
 }
 
+// refreshTouchedMCPEnvironments pushes a fresh AgentID scope list into every given
+// environment, so an already-running pod gets rolled with the new scopes right away
+// rather than only picking them up on its next deploy/promote. Best-effort: the MCP
+// config change itself already succeeded, so a refresh failure here must never turn
+// that success into an error response — it's logged and the agent simply picks up
+// the change on its next deploy/promote/rotation instead.
+func (s *agentConfigurationService) refreshTouchedMCPEnvironments(ctx context.Context, ouID, projectName, agentName string, touchedEnvNames map[string]struct{}) {
```
jhivandb, line 2437: **Unconditional pod restarts on every MCP config update.** `InjectForEnvironment` always stamps `restartedAt` and rolls the pod, and `touchedEnvNames` is the union of *all existing* and *all requested* env mappings — so any MCP config update (including a no-op) restarts pods in **every** mapped environment, even ones whose binding didn't change.

The PR already contains the right tool: `ReconcileForEnvironment` compares the desired scope list against the live workload and skips the write when in sync. Calling that here instead makes the refresh free for unchanged environments.

agent-manager-service/services/mcp_proxy_service.go
```diff
@@ -467,12 +476,60 @@ func (s *MCPProxyService) Update(ctx context.Context, orgUUID, proxyID string, r
...
+
 	return convertModelMCPProxyToSpec(updated), nil
 }
 
+// refreshAgentsBoundToProxy re-injects AgentID credentials (recomputing the scope list)
+// for every agent bound to proxyUUID, across every environment they're deployed to. Purely
+// best-effort: this runs after the proxy update already succeeded, so a failure here must
+// never surface as an error from Update — it's logged and the corresponding agent simply
+// picks up the change on its next deploy/promote/rotation instead.
+func (s *MCPProxyService) refreshAgentsBoundToProxy(ctx context.Context, proxy *models.MCPProxy, orgUUID string) {
```
jhivandb, line 519: **Restart storm: any proxy update restarts every bound agent in every environment.** This fires on *any* `MCPProxyService.Update` — editing a proxy's description or URL rolls the pods of every agent bound to it, org-wide, even when `ToolScopeBindings` didn't change.

Same fix as the config-service hook: call `ReconcileForEnvironment` (which no-ops when the scope list is unchanged) instead of `InjectForEnvironment` (which unconditionally stamps `restartedAt`). Alternatively, only trigger this when the update actually touched `ToolScopeBindings`.

agent-manager-service/services/agent_thunder_reconciler.go
```diff
@@ -158,4 +158,31 @@ func (s *agentThunderReconcilerService) runCycle(ctx context.Context) {
...
+
+	s.runIdentityInjectionReconcile(ctx)
+}
+
+// identityInjectionReconcileWindow bounds how far back the injection sweep
+// looks. It only bridges the gap between provisioning completing (seconds
+// after create) and a brand-new git agent's first build creating its
+// ReleaseBinding (minutes later); steady-state sync is owned by the deploy,
+// promote, rotation, and MCP-config-change paths, so the window is finite.
+const identityInjectionReconcileWindow = 2 * time.Hour
```
jhivandb, line 188: **Upgrade gap this window can't heal — possible cross-environment credential inheritance.** Environments promoted *before* this PR ships have ReleaseBindings with no identity overrides. After upgrade, the next redeploy to the lowest environment writes that environment's identity vars into the **shared Workload CR** — and per this PR's own comments, workload-level vars are "inherited by every environment unless a ReleaseBinding overrides it". So pre-existing staging/prod pods pick up the lowest environment's `client_id`/secret ref on their next restart. That's exactly the leak class the hard block in `PromoteAgent` exists to prevent — but the hard block only covers *future* promotions.

The 2h `created_at` window means this sweep will never touch those older bindings. Since `ReconcileForEnvironment` is already cheap when in sync, consider dropping the window (or doing a one-time backfill that asserts per-env overrides for all completed internal bindings on upgrade).

Smaller note on the query itself: oldest-first with `limit` and no pagination means a handful of stuck bindings (e.g. a build that never completes) can occupy the whole batch and starve newer agents for up to 2 hours.

jhivandb (commented):
> 🤖 **Note: this is an AI-generated review** (Claude, run by @jhivandb). Findings were verified against the checked-out branch (build + `make test-unit` pass locally), but weigh them accordingly.

## Overview

The overall shape is sound: the secret value never transits AMS on the injection path (SecretKeyRef only), the promotion hard-block direction is right (refuse rather than leak), CR naming is collision-safe, and the test coverage exercises real behavior (skip states, conflict fallback, rotation uniqueness, promotion isolation). The `DeleteAllBindings` change — continuing external cleanup after a DB delete failure — is a real fix; an orphaned Thunder identity mints tokens forever, stale rows don't.

The issues are at integration boundaries, detailed in the inline comments. Summary, by severity:

**Critical**
1. `PromoteAgent` panics when `opts.AgentThunderProvisioning` is nil — the new pre-promote call is the only unguarded call site in the file. Even with a guard, the hard block would then permanently block *all* promotions in such deployments. (inline @ `agent_manager.go:3379`)

**Architecture — OpenBao coupling**
2. The injection service reads `binding.SecretRefPath` and feeds it to `CreateSecretReferenceRequest.KVPath` ("Path in OpenBao KV store"), hardcoding OpenBao semantics *outside* the deployment seam (`app.Options.AgentThunderProvisioning`). A non-OpenBao provisioning implementation compiles fine and breaks silently at runtime. Suggest a `CredentialLocation(binding)` method on the provisioning interface so only the provisioning impl interprets that column. (inline @ `agent_identity_injection_service.go:384`)
3. `SetWorkloadInjector`/`ReconcileWorkloadInjection` pollute the deployment-seam interface to work around construction order. (inline @ `agent_thunder_provisioning_service.go:139`)

**High**
4. Both MCP refresh hooks call `InjectForEnvironment` (always rolls the pod) instead of `ReconcileForEnvironment` (skips when in sync) — any proxy update restarts every bound agent in every environment. (inline @ `mcp_proxy_service.go:495`, `agent_configuration_service.go:2427`)
5. The promotion hard block conflates four distinct nil-states; its error message is wrong for revoked/FAILED bindings, and first promote to a new environment always fails. (inline @ `agent_manager.go:3401`)
6. Upgrade gap: environments promoted *before* this PR have ReleaseBindings without identity overrides; the next redeploy to the lowest environment writes identity vars into the shared Workload CR, and those pre-existing pods inherit the lowest environment's credentials on their next restart. The 2h reconciler window can't heal them. (inline @ `agent_thunder_reconciler.go:170`)

**Medium/minor** — silent best-effort rotation refresh, swallowed scope-resolution errors causing rollout flapping, redundant duplicate `ProvisionForEnvironmentIfMissing`, retry helper ignoring ctx/permanent errors, unconditional `UpdateSecretReference` write amplification. All inline.

Also: the PR description says "scopes are a fixed placeholder today", but the code fully resolves scopes from MCP tool-scope bindings — one of the two is stale.

**Verdict: request changes** — #1 is a crash in a supported configuration, #4 is a two-line fix with fleet-wide operational impact, and #2/#3 get much more expensive to fix after more consumers appear.

agent-manager-service/services/agent_manager.go
```diff
@@ -3095,6 +3122,18 @@ func (s *agentManagerService) RegenerateAgentIdentitySecret(ctx context.Context,
 		return nil, err
 	}
 
+	// Gateway Binding: an internal agent's running pod still holds the OLD
+	// secret via its SecretReference-backed env var. Force an immediate Secret
+	// re-sync + pod rollout so it picks up the rotated value. Best-effort —
+	// the rotation itself already succeeded, so the response must report it
+	// regardless; a failed refresh self-heals on the next deploy.
+	if ownership == models.AgentProvisioningTypeInternal {
+		if refreshErr := s.agentIdentityInjection.RefreshAfterRotation(ctx, ouID, projectName, agentName, environmentName); refreshErr != nil {
```
jhivandb, line 3188: **Silent failure mode: pod keeps an invalidated secret indefinitely.** If this refresh fails, the rotation has already invalidated the old secret in Thunder, but the pod's env var (from `SecretKeyRef`) won't update without a restart — even the SecretReference's periodic re-sync only updates the K8s Secret, not the running pod's environment. "Self-heals on the next deploy" in practice means "the agent is broken until someone deploys", and the API response gives the caller no indication anything went wrong.

Consider surfacing a warning field in `AgentRegenerateSecretResponse` when the refresh fails, so the caller knows the workload hasn't picked up the new credential yet.

agent-manager-service/services/agent_identity_injection_service.go
```diff
@@ -0,0 +1,612 @@
...
+// DB, no Thunder role/group lookups. Requesting a scope the AgentID isn't
+// actually authorized for is safe: Thunder filters requested scopes down to
+// what the agent's role assignments actually grant at token-mint time, so
+// this is a "what might I need" list, not the enforcement point.
+//
+// Fails closed: any lookup error (environment resolution, agent config load)
+// logs a warning and returns nil (no scopes) rather than a stale or
+// over-broad set. A transient DB/OpenChoreo blip costing an agent temporary
+// MCP access is safe; silently keeping the wrong scopes is not.
+func (s *agentIdentityInjectionService) resolveAgentIdentityScopes(ctx context.Context, binding *models.AgentThunderClient) []string {
```
jhivandb: **Swallowed scope-resolution errors cause rollout flapping and silent scope loss.** A transient DB/OpenChoreo blip here returns nil scopes with only a warn log, so:

- In the deploy path, the deploy proceeds and injects an **empty** `AMP_AGENT_IDENTITY_SCOPES` — inconsistent with the same flow, where a binding-load error aborts the deploy to "prevent credential loss".
- In the reconcile path, one blip makes `desired` scopes empty ≠ current → a pod rollout onto empty scopes, then a second rollout back once the blip clears — contradicting `ReconcileForEnvironment`'s "never causes a needless pod rollout" contract.

Since Thunder already filters requested scopes at token-mint time (as this comment notes), the "fail closed" rationale doesn't buy safety — returning the error is strictly better; every caller already handles errors.

agent-manager-service/services/agent_identity_injection_service.go
```diff
@@ -0,0 +1,612 @@
...
+// next unrelated deploy/promote/rotation happens to re-assert it.
+const (
+	releaseBindingUpdateRetries    = 3
+	releaseBindingUpdateRetryDelay = 500 * time.Millisecond
+)
+
+// withReleaseBindingRetry retries fn a bounded number of times, sleeping
+// between attempts. Every call site here already wraps a single, cheap,
+// idempotent Get-then-Update — retrying is always safe.
+func withReleaseBindingRetry(fn func() error) error {
```
jhivandb: Two issues with the retry helper:
- No `ctx` awareness — `time.Sleep` runs out its budget even when the request is cancelled. Take a `ctx` and use a timer/select.
- It retries **every** error, including permanent ones (404, 403, validation), adding 1s of pointless latency to genuine failures. The stated motivation is optimistic-concurrency conflicts — retry only on conflict.

agent-manager-service/services/agent_identity_injection_service.go
```diff
@@ -0,0 +1,612 @@
...
+		return nil, nil //nolint:nilnil // credential revoked — nothing valid to inject
+	}
+	return binding, nil
+}
+
+// ensureSecretReference creates or updates the SecretReference CR that lets
+// OpenChoreo materialize the OpenBao credential as a Kubernetes Secret.
+// templateAnnotations may be nil; see secretRotatedAtAnnotation for when it
+// is not.
+func (s *agentIdentityInjectionService) ensureSecretReference(ctx context.Context, binding *models.AgentThunderClient, templateAnnotations map[string]string) (string, error) {
```
jhivandb, line 360: **Write amplification + annotation churn.** Every `EnvVarsForEnvironment` call lands here and does an unconditional `UpdateSecretReference` — that's one k8s API write per deploy/promote/config-update *and per reconciler tick per recent binding*. Two side effects:
- After a rotation stamped `secret-rotated-at`, the next non-rotation update carries nil `TemplateAnnotations` and (assuming replace semantics) strips the annotation — a spec change that triggers an extra controller re-sync each way.
- The reconciler's "no needless writes" property holds for pod rollouts but not for this CR.

Consider comparing the existing SecretReference against the desired request and skipping the update when unchanged.

agent-manager-service/services/agent_thunder_reconciler.go
```diff
@@ -158,4 +158,31 @@ func (s *agentThunderReconcilerService) runCycle(ctx context.Context) {
...
+// case is a duplicate content-identical write.
+func (s *agentThunderReconcilerService) runIdentityInjectionReconcile(ctx context.Context) {
+	createdAfter := time.Now().Add(-identityInjectionReconcileWindow)
+	recent, err := s.repo.FindRecentlyCompletedInternal(ctx, createdAfter, reconcilerBatchSize)
+	if err != nil {
+		s.logger.Error("Failed to query recently completed internal agent thunder bindings", "error", err)
+		return
+	}
+	for _, binding := range recent {
+		s.provisioning.ReconcileWorkloadInjection(ctx, binding)
```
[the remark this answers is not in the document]
yashed: Fixed: the reconciler now pages through every eligible record each cycle instead of repeatedly grabbing only the oldest 50, so newer agents can no longer be starved out of getting their credentials attached.

agent-manager-service/services/agent_thunder_provisioning_service.go
```diff
@@ -631,11 +701,28 @@ func (s *agentThunderProvisioningService) DeleteAllBindings(ctx context.Context,
...
 
 	for _, b := range bindings {
+		// The AgentID SecretReference CR is independent of whether the identity
+		// ever landed in Thunder — clean it up for every internal binding, even
+		// ones that never completed.
+		if s.workloadInjector != nil && b.ProvisioningType == models.AgentProvisioningTypeInternal {
+			if err := s.workloadInjector.CleanupForEnvironment(ctx, ouID, agentName, b.EnvironmentName); err != nil {
+				s.logger.Warn("Failed to delete agent identity SecretReference during agent deletion", "agentName", agentName, "env", b.EnvironmentName, "error", err)
+			}
+		}
```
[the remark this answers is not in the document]
yashed: Fixed: the record's pointer to its credentials is now cleared the moment deletion starts, so even if the database row fails to delete, the leftover record can never be mistaken for a healthy one later.

agent-manager-service/services/agent_thunder_provisioning_service.go
```diff
@@ -289,6 +343,11 @@ func (s *agentThunderProvisioningService) ProvisionForEnvironmentIfMissing(
 }
 
 func (s *agentThunderProvisioningService) AttemptProvision(ctx context.Context, binding models.AgentThunderClient) {
+	// Held for the entire attempt so this can't interleave its Thunder
+	// RegenerateAgentSecret/OpenBao Store calls with a concurrent
+	// RegenerateSecret/RevokeSecret on the same binding.
+	defer s.bindingLocks.Lock(bindingLockKey(binding.OUID, binding.ProjectName, binding.AgentName, binding.EnvironmentName))()
+
```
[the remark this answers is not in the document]
yashed: Addressed with a smaller fix: since the underlying calls already have hard timeouts, made the lock cancellable so a waiting request gives up as soon as its own deadline hits, instead of a bigger redesign.

agent-manager-service/services/agent_thunder_provisioning_service.go
```diff
@@ -129,27 +129,77 @@ type AgentThunderProvisioningService interface {
...
+	s.workloadInjector = injector
+}
+
+func (s *agentThunderProvisioningService) ReconcileWorkloadInjection(ctx context.Context, binding models.AgentThunderClient) {
+	if s.workloadInjector == nil || binding.ProvisioningType != models.AgentProvisioningTypeInternal {
+		return
+	}
+	if err := s.workloadInjector.ReconcileForEnvironment(ctx, binding.OUID, binding.ProjectName, binding.AgentName, binding.EnvironmentName); err != nil {
+		s.logger.Warn("Failed to reconcile agent identity credentials into workload",
+			"agentName", binding.AgentName, "envName", binding.EnvironmentName, "error", err)
```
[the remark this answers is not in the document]
yashed: Fixed: added the organization ID and record ID, both already available in the code, to all 7 affected log lines so failures can be traced.

agent-manager-service/services/agent_thunder_provisioning_service.go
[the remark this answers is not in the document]
yashed: Addressed with a smaller fix instead of the full rewrite: kept the existing setup order but added a warning log so a future setup mistake would be immediately visible instead of silently failing.

agent-manager-service/services/agent_manager.go
```diff
@@ -3290,6 +3348,65 @@ func (s *agentManagerService) PromoteAgent(ctx context.Context, ouID string, pro
...
+	// lines above for LLM/system configuration (len(tgtSystemKeys) == 0 case).
+	if len(tgtIdentityEnvVars) == 0 {
+		return fmt.Errorf(
+			"%w: agent %q's AgentID identity for target environment %q is not ready yet (still provisioning). "+
+				"Promotion is blocked to prevent the promoted pod from inheriting a different environment's credentials — "+
+				"check GET .../identities for this agent and retry once it shows status=completed for %q",
+			utils.ErrInvalidInput, agentName, req.TargetEnvironment, req.TargetEnvironment,
+		)
+	}
+
```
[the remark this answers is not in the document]
yashed: Fixed: wrapped the provisioning call and its readiness check in a nil guard so promotions still work normally on deployments where this feature isn't enabled.

agent-manager-service/services/agent_identity_injection_service.go
```diff
@@ -0,0 +1,612 @@
...
+			time.Sleep(releaseBindingUpdateRetryDelay)
+		}
+	}
+	return lastErr
+}
+
+// agentIdentitySecretRefKind distinguishes the AgentID SecretReference name
+// from the "<entity>-<env>-secrets" names secretmanagersvc generates for
+// LLM/MCP/user secrets, so the two families can never collide.
+const agentIdentitySecretRefSuffix = "agent-identity"
```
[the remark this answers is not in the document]
yashed: Fixed

agent-manager-service/services/agent_identity_injection_service.go
```diff
@@ -0,0 +1,612 @@
...
+	for attempt := 0; attempt < releaseBindingUpdateRetries; attempt++ {
+		if lastErr = fn(); lastErr == nil {
+			return nil
+		}
+		if attempt < releaseBindingUpdateRetries-1 {
+			time.Sleep(releaseBindingUpdateRetryDelay)
+		}
+	}
+	return lastErr
+}
```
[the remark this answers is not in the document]
yashed: Fixed: the retry now only retries actual write conflicts and stops immediately if the caller's own request is cancelled, instead of blindly retrying every kind of error for the full window.

agent-manager-service/services/agent_configuration_service.go
yashed: Fixed: swapped InjectForEnvironment for ReconcileForEnvironment in refreshTouchedMCPEnvironments, so environments whose scopes didn't actually change no longer get rolled.

agent-manager-service/services/mcp_proxy_service.go
yashed: Already covered by the same fix as the config-service hook above

agent-manager-service/services/agent_manager.go
yashed: Fixed: added a `workloadRefreshWarning` field to the regenerate-secret response that's populated whenever the automatic pod refresh fails, so the caller knows the workload hasn't picked up the new credential yet.

agent-manager-service/services/agent_identity_injection_service.go
yashed: Fixed : `resolveAgentIdentityScopes` now returns an error instead of silently falling back to an empty scope list, since every caller already aborts correctly on error and an empty scope list was never actually safer.

agent-manager-service/services/agent_identity_injection_service.go
yashed: Fixed: ensureSecretReference now skips the update entirely when the existing SecretReference already points at the desired KV path and keys, so steady-state calls stop rewriting it and a prior rotation's annotation is never silently clobbered.

agent-manager-service/services/agent_identity_injection_service.go
yashed: Already fixed, `withReleaseBindingRetry` already takes a context (returns immediately on cancellation instead of sleeping out the delay) and only retries on the optimistic-concurrency conflict, not permanent errors.

agent-manager-service/services/agent_thunder_reconciler.go
yashed: Fixed — added a one-time, unwindowed reconcile backfill that runs on every service start

agent-manager-service/services/agent_manager.go
yashed: fixed

jhivandb (commented):
## 🤖 AI code review

Automated review of the AgentID credential injection changes. Each finding below was independently verified against the code (adversarial pass — I tried to refute each before keeping it). Line numbers are approximate; anchors point at real code. Two lower-confidence items from the first draft were dropped after verification (annotation-wipe was unreachable under the code's own invariants; the `SetWorkloadInjector` ordering concern is already covered by unit tests + a self-healing reconciler).

---

### 🔴 High — Cross-environment credential leak on promote when provisioning is disabled

`services/agent_manager.go` — `PromoteAgent`

The deploy path writes the **lowest** environment's real `client_id`/`client_secret` into the shared OpenChoreo Workload CR, which every environment inherits unless its own ReleaseBinding overrides them. `PromoteAgent` is supposed to override the identity keys on the target binding (`tgtIdentityEnvVars`) and hard-block if the target's identity isn't ready — but that **entire** block (kickoff, `EnvVarsForEnvironment`, `pollForTargetIdentityReady`, hard block, and the `tgtIdentityEnvVars` append) lives inside `if s.agentThunderProvisioning != nil`. When that field is nil, `tgtIdentityEnvVars` stays empty and nothing overrides the identity keys on the target.

`PromoteComponent` (`clients/openchoreosvc/client/deployments.go`) does `binding.Spec.WorkloadOverrides = workloadOverrides` — a **full replace**, not a merge — so the identity keys simply fall through to the shared Workload CR base.

The asymmetry that makes this reachable: the **deploy path's** identity injection (`agentIdentityInjection.EnvVarsForEnvironment`, wired unconditionally) is **not** gated on `agentThunderProvisioning`. So a deployment that provisioned bindings while enabled, then runs with provisioning disabled, keeps writing the lowest env's real creds into the Workload CR while promote silently omits the target override → the promoted pod inherits the lowest env's real `client_id`/`client_secret`. No error, no warning.

**Fix:** run `EnvVarsForEnvironment` + the hard block regardless of `agentThunderProvisioning` (only the `ProvisionForEnvironmentIfMissing` kickoff/poll genuinely needs it), mirroring the ungated deploy path; or explicitly refuse to promote internal agents when provisioning is off.

---

### 🟠 Medium — `withReleaseBindingRetry` never actually retries on a conflict

`services/agent_identity_injection_service.go` (helper, ~L82) → `InjectForEnvironment`, `RefreshAfterRotation`, `RemoveForEnvironment`

The retry gate is `if !errors.Is(lastErr, utils.ErrConflict) { return lastErr }`, but the wrapped OpenChoreo calls never produce `ErrConflict`:
- `ErrConflict` is only ever built from `JSON409` in `errors.go`.
- The generated `UpdateReleaseBindingResp` / `UpdateWorkloadResp` types have **no `JSON409` field at all** — so it can't be mapped even if the `ErrorResponses{}` literals were "fixed".
- Per this repo's own sibling retry loop (`deployments.go` ~L265-268), OpenChoreo currently surfaces conflicts as HTTP **500**, not 409 — which maps to `ErrInternalServerError`, not `ErrConflict`.

So the retry is effectively dead: a real concurrency conflict returns on the first attempt. (Note: the error *does* surface to the caller — it is not silently swallowed — so impact is a lost auto-recovery, not silent data loss. Downgraded from the original High on that basis.) The unit test passes only because it mocks the wrapped func to return `utils.ErrConflict` directly, bypassing the real error-construction path.

**Fix:** the "add `JSON409`" fix won't work (no such field, and conflicts are 500s). Retry on `StatusCode() == 409 || == 500` at the call site, mirroring `deployments.go`.

---

### 🟠 Medium — `DeleteAllBindings` takes no per-binding lock → orphaned live Thunder identity

`services/agent_thunder_provisioning_service.go` — `DeleteAllBindings`

`AttemptProvision`, `RegenerateSecret`, and `RevokeSecret` all serialize per-binding work via `s.bindingLocks.Lock(...)`. `DeleteAllBindings` does not — it snapshots rows via `FindByAgent` then `DeleteByIDs`. `UpdateAfterAttempt` (`repositories/agent_thunder_client_repository.go`) runs `...Updates(updates).Error` and does **not** check `RowsAffected`.

Race: a `DeleteAgent` request races an in-flight `AttemptProvision` (reconciler sweep / detached goroutine) for the same binding. `DeleteByIDs` removes the row; `AttemptProvision` then finishes creating a real Thunder identity + OpenBao secret and calls `UpdateAfterAttempt`, which matches 0 rows and returns `nil`. The Thunder identity is now orphaned — no DB row references it, so nothing cleans it up, and it can keep minting valid tokens. `DeleteAllBindings`'s own Thunder-delete loop can't catch it either: the snapshot was taken before the attempt set `ThunderAgentID`, so `if b.ThunderAgentID == "" { continue }` skips it. (The existing "crash orphan is a harmless tradeoff" comment covers only a crash window, not this lock-free concurrent-attempt window.)

**Fix:** acquire the per-binding lock in `DeleteAllBindings`; and/or have `UpdateAfterAttempt` signal "row gone" (check `RowsAffected`) so a completing attempt can roll back the external identity it just created.

---

### 🟠 Medium — Revoking the lowest env during a pipeline-fetch error leaves a dangling `SecretKeyRef`

`services/agent_manager.go` — `RevokeAgentIdentitySecret` (~L3173-3180)

```go
includeWorkloadLevel := false
if pipeline, pipeErr := s.ocClient.GetProjectDeploymentPipeline(...); pipeErr != nil {
    s.logger.Warn(...) // includeWorkloadLevel stays false
} else {
    includeWorkloadLevel = findLowestEnvironment(pipeline.PromotionPaths) == environmentName
}
```

`includeWorkloadLevel` gates whether `RemoveForEnvironment` also strips the identity vars from the shared Workload CR (the `RemoveWorkloadEnvVars` call is inside `if includeWorkloadLevel`). But `RemoveForEnvironment` calls `deleteSecretReference` **unconditionally** (outside that guard). So revoking the lowest env while `GetProjectDeploymentPipeline` transiently errors → `includeWorkloadLevel` stays false → the Workload-level vars are retained (they still contain a `SecretKeyRef` pointing at the per-env CR) but the CR itself is deleted → a future/restarting lowest-env pod fails with `CreateContainerConfigError`. The revoke response has no field to signal this (unlike the new `WorkloadRefreshWarning` on regenerate), so the caller sees plain success.

**Fix:** don't delete the SecretReference CR while workload-level vars still reference it — on pipeline-resolution failure, either fail the revoke closed, or defer the CR deletion, rather than proceeding with `includeWorkloadLevel = false` when it might be the lowest env.

---

### 🟡 Low — `RevokeSecret`/`ClaimSecret` clear `SecretRefPath` even when the delete failed

`services/agent_thunder_provisioning_service.go` — `RevokeSecret` (~L784), `ClaimSecret` (~L878)

```go
if binding.SecretRefPath != "" {
    if err := s.deleteCredential(...); err != nil {
        s.logger.Warn("Failed to delete stored secret during revoke", ...) // non-fatal: "can be retried by revoking again"
    }
}
if err := s.repo.UpdateSecretRef(ctx, binding.ID, ""); err != nil { // runs regardless
    return "", err
}
```

`UpdateSecretRef(..., "")` runs even when `deleteCredential` just failed. On the next revoke, `binding.SecretRefPath == ""`, so the `if binding.SecretRefPath != ""` guard is false and `deleteCredential` is never retried — the comment promising retry is wrong. A transient OpenBao/OpenChoreo error therefore orphans the KV secret **and** its SecretReference CR permanently. (Pre-existing pattern; the swap from `secretStore.Delete` to `deleteCredential` re-exposes it and now leaks the CR too, not just the KV entry. The leaked KV points at an already-invalidated Thunder secret, so no valid credential leaks — storage/CR leak only, hence Low.)

**Fix:** only clear `SecretRefPath` when `deleteCredential` succeeds; otherwise leave it so a re-revoke retries.

---

### 🟡 Low — `readCredential` returns empty id/secret without error → burnable one-time claim

`services/agent_thunder_provisioning_service.go` — `readCredential` (~L243)

```go
data, err := s.secretMgmtClient.GetSecretWithValue(ctx, secretRefPath)
if err != nil { return "", "", err }
return data[AgentSecretKeyClientID], data[AgentSecretKeyClientSecret], nil
```

The deleted store's `Get()` mapped "both values empty" to `ErrAgentSecretNotFound`; this returns `("", "", nil)`. `ClaimSecret` does its one-time `MarkClaimed` CAS **before** the read and only rolls back (`ClearClaim`) when `readCredential` returns a non-nil error. So if `GetSecretWithValue` succeeds but the map lacks the expected keys (malformed/partial write), the one-time claim is consumed, no rollback fires, and `ClaimSecret` returns success with an empty `client_secret` — no retry path. Clean regression of the old not-found behavior.

**Fix:** return `ErrSecretNotFound` (or an error) from `readCredential` when both values are empty, so `ClaimSecret`'s existing rollback re-engages.

---

### 🟡 Low — Synchronous, unbounded request-path fan-out on MCP updates

`services/mcp_proxy_service.go` (~L499/537), `services/agent_configuration_service.go` (~L2416)

`MCPProxyService.Update` calls `refreshAgentsBoundToProxy` **synchronously** on the request goroutine. It loops over every (agent, env) mapping bound to the proxy and, per iteration serially, runs `agentConfigRepo.GetByUUID` (7 eager preloads, only `ProjectName`/`AgentID` used — and `ReconcileForEnvironment` reloads the full config again internally via `GetByAgentID`) + `ReconcileForEnvironment` (which itself makes ~4+ OpenChoreo round-trips even on the in-sync no-op path: `GetSecretReference` + `Create`/`UpdateSecretReference` + `GetEnvironment` + `GetComponentConfigurations`). Cost scales with agents × environments, serially, inside the request → gateway-timeout risk for a popular shared proxy. `updateMCPConfig → refreshTouchedMCPEnvironments` has the same shape but is bounded to one agent's env count.

**Fix:** detach the refresh onto a background goroutine (it's already best-effort) — but **use a derived/background context**, not the request `ctx`, which is cancelled when `Update` returns. Replace the per-mapping `GetByUUID` with a lightweight `uuid, project_name, agent_id` projection. Optionally skip entirely when `ToolScopeBindings` didn't change, and avoid the `GetSecretReference` + `Create`/`Update` writes on the already-in-sync path.

---

### 🟡 Low (latent) — Path-traversal defense-in-depth dropped for the AgentID credential flow

`services/agent_thunder_provisioning_service.go` — `agentIdentitySecretLocation` (~L208) → relies on `secretmanagersvc.sanitizeSegment`

The deleted `agent_secret_store.go`'s `key()` rejected `""`, `.`, `..`, and `/` in every segment. The replacement path helpers do no such pre-check; the only guard is `sanitizeSegment`, which rejects `/` only — `.` and `..` pass through to `path.Join`/`path.Clean` in the OpenBao provider, which collapse them.

**Not currently exploitable:** every credential *store*/*delete* path is preceded by `env_resolver.Resolve`, which re-added the full `""`/`.`/`..` rejection for org/env; and `projectName`/`agentName` are validated to `[a-z0-9-]+` upstream. So a `..` segment fails at `Resolve` before reaching the store. This is a defense-in-depth regression (latent fragility if a future store path skips `Resolve`), not a live hole — hence Low.

**Fix:** reject empty/`.`/`..`/`/` segments in `agentIdentitySecretLocation` inputs, or centralize the check in `sanitizeSegment`.

---

### ⚪ Nit — `workloadInjector` read at ~L945 bypasses the mutex getter

`services/agent_thunder_provisioning_service.go` (~L945)

`DeleteAllBindings` reads `s.workloadInjector` directly (`if s.workloadInjector != nil && ...`) instead of via `getWorkloadInjector()` (RLock) like every other reader. **This is not a data race** — the startup ordering in `app.go` writes the field on the main goroutine before the reconciler/HTTP-server goroutines are created, establishing happens-before, so no reader observes it concurrently. But it's an inconsistency with the field's documented invariant, and unlike `reconcileWorkloadInjection` this path silently no-ops (no warn) when the injector is nil.

**Fix:** route through `s.getWorkloadInjector()` into a local, and log when nil.

---

### ⚪ Nit — `AgentIdentityEnvVarKeys()` allocates a fresh map per call; system-managed boilerplate duplicated

`services/agent_identity_injection_service.go` (~L101) and 3 sites in `agent_manager.go` (deploy/promote/config-update)

The function returns a freshly-allocated 4-entry `map[string]bool` on every call (`RemoveForEnvironment` calls it twice back-to-back), and the "ensure map non-nil, then mark these keys system-managed" block is copy-pasted across deploy/promote/config-update. A future 5th key must be updated in four places.

**Fix:** a package-level read-only `agentIdentityEnvVarKeys` var + one helper that merges it into a caller's system-managed set.

---

### ⚪ Nit — Redundant `GetEnvironment`/binding re-fetches per deploy/promote/config-update

`services/agent_identity_injection_service.go` — `resolveAgentIdentityScopes` (~L253)

`resolveAgentIdentityScopes` calls `ocClient.GetEnvironment(ouID, envName)` (when the agent has MCP mappings — it early-returns otherwise) to resolve the environment UUID, but callers already resolved that same `Environment` in the same request (config-update validates via `GetEnvironment` ~18 lines earlier; deploy/promote similarly). `PromoteAgent` also does `repo.Get` in `ProvisionForEnvironmentIfMissing` immediately followed by another `repo.Get` inside `EnvVarsForEnvironment`.

**Fix:** let `resolveAgentIdentityScopes`/`EnvVarsForEnvironment` accept the caller's already-resolved environment (or UUID) as a parameter. (Requires an interface signature change, so a deliberate call — flagging as efficiency only.)

agent-manager-service/services/agent_thunder_provisioning_service_test.go
```diff
@@ -1168,34 +1328,180 @@ func TestDeleteAllBindings_DeletesRowsBeforeSlowThunderCleanup(t *testing.T) {
...
+	case <-deleteDone:
+	case <-time.After(2 * time.Second):
+		t.Fatal("DeleteAllBindings never completed after the in-flight attempt released the lock")
+	}
+
+	assert.Equal(t, []string{"thunder-agent-1"}, deletedThunderIDs,
+		"the Thunder identity the concurrent attempt just created must be cleaned up, not orphaned")
+	assert.Equal(t, []string{"dev"}, deletedSecretEnvs,
+		"the OpenBao secret the concurrent attempt just stored must be cleaned up, not orphaned")
 }
```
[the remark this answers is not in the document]
yashed: Releasing the lock early like suggested could accidentally delete a brand-new secret, so instead each binding now unlocks as soon as its own cleanup is done instead of waiting for all of them — tested and confirmed working.

jhivandb (approved):

PR #1395, opened 2026-07-23
Correct the simple VM path's exposed hosts and teardown

RAVEENSR:
## Purpose
Three statements in the **Simple** tab of the "Run Agent Manager on a VM with Docker" guide no longer describe what the installer does. All three were found by following the page end to end on a clean Ubuntu VM and comparing each claim against the resulting install.

1. **Teardown cannot be executed as written.** The page says to run `cd agent-manager/deployments/quick-start && sudo ./uninstall.sh --delete-cluster`. The simple installer downloads a versioned bundle rather than cloning the repository, and `bootstrap.sh` unpacks that bundle into a `mktemp -d` which its `EXIT` trap removes, so no checkout and no `uninstall.sh` exist on the VM afterwards. A reader following the teardown section has nothing to `cd` into.
2. **The OTel ingest host is attributed to the wrong agents.** `gateway.amp.<IP>.sslip.io/otel` is listed as "OTel trace ingest from deployed agents". Agents deployed by the platform run inside the cluster and export to the in-cluster gateway runtime; the public host exists for externally hosted agents that set `AMP_OTEL_ENDPOINT` themselves.
3. **Per-environment Thunder is undocumented.** The installer has always written a `*.thunder.<IP>.sslip.io` wildcard site into the Caddyfile, one host per org/environment created as environments are added. It appears in neither the exposed-hosts table nor the routing diagram, even though the Advanced tab documents the equivalent tier.

Related to #1388, which does the same accuracy pass over the Advanced tab of this page.

## Goals
Make the Simple tab's exposed-host inventory, routing diagram, and teardown instructions match the installer, so that a reader can complete both install and teardown from the page alone.

## Approach
- **Teardown** now deletes the cluster directly with `sudo k3d cluster delete amp-local` — which is exactly what `uninstall.sh --delete-cluster` did via its fast path — followed by the existing Caddy container and volume removal, plus the generated `/opt/amp/Caddyfile`. Added an explicit note that the installer leaves no checkout behind, so a reader who goes looking for an uninstall script knows why there isn't one.
- **OTel host** row now reads "OTel trace ingest for **externally-hosted** agents", with an admonition explaining that platform-deployed agents export in-cluster and never traverse Caddy, linking to the instrumentation reference where `AMP_OTEL_ENDPOINT` is documented.
- **Per-environment Thunder** added as a row in the exposed-hosts table and as a node in the routing diagram, with its own labelled Caddy → control-plane-kgateway edge. The prose above the table now names both dynamic tiers and explains once why they need on-demand TLS instead of an up-front wildcard, and the existing first-request certificate-error note is extended to cover new environments as well as new agents.

## User stories
As an operator evaluating Agent Manager on a VM, I can see every hostname the install exposes and can tear the install down again without hunting for a script the installer never placed on the machine.

## Release note
N/A — documentation only, no product change.

## Documentation
This PR is the documentation change: `documentation/docs/getting-started/on-a-vm.mdx`.

## Training
N/A — no training content covers the VM installation guide.

## Certification
N/A — no certification exam covers the VM installation path.

## Marketing
N/A — a correction to existing installation documentation, not a promotable feature.

## Automation tests
 - Unit tests
   > N/A — documentation only, no code paths changed.
 - Integration tests
   > N/A — documentation only. Verification was manual, against a live install; see **Test environment**.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A — no code changed.

## Samples
N/A — this page is itself the walkthrough; no separate sample accompanies it.

## Related PRs
- #1388 — the same accuracy pass over the Advanced tab of this page.

## Migrations (if applicable)
N/A — documentation only; nothing to migrate.

## Test environment
Every claim below was checked against a clean simple install, not a repaired one.

- **VM:** Ubuntu 26.04 LTS, 4 vCPU / 8 GB RAM / 58 GB disk, public IP, inbound `443/tcp` only.
- **Host tooling** installed by following the page's own prerequisite links: Docker 29.6.2, k3d v5.9.0, kubectl v1.36.3, Helm v3.21.3, plus `lsof`.
- **Install:** `bootstrap.sh simple` completed with exit 0 in ~11 minutes (the page estimates 8–15). 52/52 pods `Running`, none failing.
- **Exposed hosts:** all six fixed hostnames served publicly-trusted Let's Encrypt certificates (`ssl_verify_result=0`) over `:443`.
- **Teardown claim:** `sudo find / -name uninstall.sh` returned no results, and neither `~/agent-manager` nor `/root/agent-manager` exists — confirming the documented path is unreachable.
- **Per-environment Thunder claim:** the generated `/opt/amp/Caddyfile` contains a `*.thunder.<IP>.sslip.io` site block proxying to `127.0.0.1:8080` with the on-demand TLS block.
- **OTel claim:** a deployed agent's pod carries `AMP_OTEL_ENDPOINT=http://api-platform-default-default-gateway-gateway-runtime.default-default:22893/otel`, and its startup log confirms it exports there; the console's `instrumentationUrl` remains the public host, which is what an externally hosted agent uses.
- **Docs build:** `npm run build` in `documentation/` succeeds with `onBrokenLinks: 'throw'`, so the new cross-reference resolves. The edited diagram was additionally rendered with `@mermaid-js/mermaid-cli` to confirm it still parses and lays out cleanly.

## Learning
The failure mode worth recording is that all three errors were invisible from the repository: each one only shows up by running the page against a machine that has never had the platform on it. The teardown break in particular is a second-order effect of a change made elsewhere — moving the installer from a git clone to a versioned bundle silently invalidated a path in the docs that nothing links the two together. Re-walking an installation guide on a clean host after the installer's delivery mechanism changes is the only reliable way to catch that class of drift.

## Summary by CodeRabbit

- **Documentation**
  - Updated the “Run Agent Manager on a VM” guide to reflect per-environment Thunder tier behavior, including expanded “What gets exposed” routing/certificate details and environment-specific Thunder hostname guidance.
  - Clarified that externally hosted agent endpoints are handled differently in the gateway/OTEL description.
  - Improved guidance on first-invocation, one-time certificate errors for newly created per-environment Thunder hosts when adding an environment.
  - Reworked Simple installer teardown instructions with an explicit command sequence and clearer notes on what’s removed vs persisted.

### DIFF
--- a/documentation/docs/getting-started/on-a-vm.mdx
+++ b/documentation/docs/getting-started/on-a-vm.mdx
@@ -27,7 +27,7 @@ Install Agent Manager on a Linux VM where Docker is the only host dependency. Pi
 <Tabs groupId="vm-installer">
 <TabItem value="simple" label="Simple (IP + automatic TLS)" default>
 
-The simple installer exposes the platform over HTTPS using [sslip.io](https://sslip.io) hostnames derived from the VM's public IP, so there's no domain registration and no client `/etc/hosts` edits.
+The simple installer exposes the platform over HTTPS using [sslip.io](https://sslip.io) hostnames derived from the VM's public IP, so there's no domain registration.
 
 ## Prerequisites
 
@@ -111,6 +111,7 @@ flowchart TB
             api["amp-api"]
             thunder["Thunder OAuth"]
             cp["Gateway control plane<br/>(external AI gateways)"]
+            envthunder["Per-environment Thunder"]
             observer["Agent Manager Observer"]
             gw["AI Gateway<br/>OTel ingest · LLM proxy"]
             agents["Deployed agents"]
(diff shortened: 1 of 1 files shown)


RAVEENSR:
Consolidated into #1388. Both PRs corrected the same file from the same round of VM testing, were each a single commit, and were waiting on the same review — so this is now the second commit on that branch rather than a separate review pass. No content dropped: the commit is cherry-picked unchanged, and #1388's description now covers both the simple and advanced tabs.

PR #1410, opened 2026-07-25
Display status.message field in trace spans

### LINKED ISSUE #1518: [Improvement]: Reason For Guardrail Failures are not showing in Trace spans

Area/Observability (Traces, metrics, logs, monitors, scores)

Aspect/Traces (Distributed traces)

### Suggested Improvement

When a trace span fails because of a guardrail, users only see a generic "✕ Failed" chip and have no way to understand the cause.

Inspecting the raw responses returned by the AMP Observer API confirms the problem: the results don't contain the reason for such failures — they only report the status as an error.

However, when we check the raw OpenSearch trace span logs , each log has a `status.message` field that contains the full details of these guardrail failures. 

The current OpenChoreo version (1.1.1) does not return the `status.message` field when exporting these results to the AMP Observer API. OpenChoreo 1.2.0 resolves this (openchoreo/openchoreo#4199) by including status.message in the observability trace API responses.

Now that 1.2.0 provides the status.message field, the AMP console trace view can display the reason for guardrail failures as well.

Expected behavior

- The amp Observer backend surfaces the span status message (alongside error) through the trace API.
- The console shows the failure reason in the span details.
- For guardrail failures, the human-readable actionReason is extracted from the payload and shown clearly (e.g. as a chip on the span header) instead of the raw JSON/dict string.

RavinduWeerakoon:
## Purpose
When observing agentic applications, error details(like guardrail failiures) are frequently stored inside status.message field inside the trace logs.
`Error code: 422 - {'message': {'action': 'GUARDRAIL_INTERVENED', 'actionReason': 'Violation of applied content length constraints detected.', 'direction': 'REQUEST', 'interveningGuardrail': 'content-length-guardrail'}, 'type': 'CONTENT_LENGTH_GUARDRAIL'}`

The current implementation of the agent manager does not have the capability to display those messages in the traces page and this pr tries to fix it

## Goals
Resolves #1518

## Approach
The current openchoreo observability API does not support the status.message field on their api spec and this pr introduces OC 1.2.0 with the update of emiting status.message field in the trace query api requests
image

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary by CodeRabbit

* **New Features**
  * Trace details now include developer-facing status messages, including structured error information.
  * Status messages containing JSON or Python-style data are formatted for easier viewing.
  * Added project type and project-environment configuration for platform resource deployments.

* **Bug Fixes**
  * Improved deployment setup by synchronizing required resource definitions before installation.
  * Installation checks now distinguish temporarily unavailable pod logs from genuine errors.

* **Configuration**
  * Updated quick-start and environment configurations for the latest platform and observability versions.


console/workspaces/pages/traces/src/subComponents/spanDetails/Overview.tsx
```diff
@@ -93,6 +93,110 @@ function formattedMessage(message: string) {
     return message;
   }
 }
+
+function pythonDictToJson(input: string): string {
```
AnoshanJ: `pythonDictToJson` and `parseEmbeddedObject` are duplicated here and in `statusMessage.ts` (identical char for char). Nothing here imports the shared module.

Was this copy left behind after the extraction? Can we export `formatStatusMessage` from `statusMessage.ts` and import it instead?

deployments/setup/setup-openchoreo.sh
```diff
@@ -29,6 +29,56 @@ echo ""
...
+        return 1
+    fi
+    if compgen -G "$tmp/${chart}/crds/*.yaml" >/dev/null; then
+        echo "   Syncing ${chart} CRDs to ${OPENCHOREO_VERSION} (helm upgrade does not update crds/)..."
+        if ! kubectl apply --server-side --force-conflicts -f "$tmp/${chart}/crds/" >/dev/null; then
+            echo "❌ Failed to apply ${chart} CRDs" >&2
+            return 1
+        fi
+    fi
+}
```
AnoshanJ, line 62: `_sync_chart_crds_body` ends on the `if compgen` block with no explicit return, so a chart with no `crds/*.yaml` returns non-zero for the normal case.

Can we add `return 0` before the closing brace?

agent-manager-observer/observer/types.go
```diff
@@ -75,7 +80,7 @@ type SpanInfo struct {
 	EndTime            time.Time              `json:"endTime"`
 	DurationNs         int64                  `json:"durationNs"`
 	Kind               string                 `json:"spanKind,omitempty"`
-	Status             string                 `json:"status,omitempty"`
+	Status             *SpanStatus            `json:"status,omitempty"`
```
AnoshanJ, line 114: `string` -> `*SpanStatus` cannot decode the old shape (`cannot unmarshal string into Go struct field SpanInfo.status`). Since `client.go:179` unmarshals the whole response at once, one string status fails the entire trace query, not just one span. So adapter 0.5.x returns zero traces, and `env.sh` notes upstream still pins 0.5.0.

Is the hard version pin intended? If not, an `UnmarshalJSON` accepting both shapes would drop the coupling.

AnoshanJ (commented):
Please check comments below.

console/workspaces/pages/traces/src/subComponents/spanDetails/Overview.tsx
RavinduWeerakoon: yeah it's a copy left behind after extraction, I have now changed it to import from the ts file

agent-manager-observer/observer/types.go
RavinduWeerakoon: yeah the version pin is intended since because 0.6.0 is the version that returns span status as an object with a message

RavinduWeerakoon: Then do we have to remove the pin and wait until the upstream update the tracing version while configuring the stuff to work in both manner or remove the pin after the oc guys update it?

AnoshanJ: Lets not remove the pin. There is no case we use this amp-observer with adapter 0.5.x, so should be fine.

AnoshanJ:
This looks good. In addition to the chip, shall we retain the 'Status' column too? @RavinduWeerakoon This will be similar to the Scores, which shows both chip and Scores column when applicable, making it easier to find the info.

console/workspaces/pages/traces/src/subComponents/spanDetails/statusMessage.ts
```diff
@@ -0,0 +1,162 @@
+/**
+ * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
```
rasika2012:
```suggestion
 * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
```

RavinduWeerakoon:
> This looks good. In addition to the chip, shall we retain the 'Status' column too? @RavinduWeerakoon This will be similar to the Scores, which shows both chip and Scores column when applicable, making it easier to find the info.

@AnoshanJ  
image

AnoshanJ (approved):

Issue #1058, opened 2026-06-12
Improve install on your environment documentation instructions


menakaj:
Improve the installation instructions in a user friendly manner.
Optimize the deployment resources

PR #1438, opened 2026-07-28
Rewrite install guide for production deployments and allow TLS on the database connection

RAVEENSR:
## Purpose

The **On Your Environment** guide is the entry point for deploying Agent Manager on a real cluster, but it was written as an install walkthrough with production advice gathered into a trailing section. That structure breaks down for settings that cannot be changed after installation.

Thunder seeds every platform OAuth client into its database through a `pre-install` hook that `helm upgrade` never re-runs. Its database, its client secrets, and its issuer/redirect hostnames are therefore frozen at first boot. A reader who followed the install steps and only then worked through the trailing "production" section would find that the guidance could not be applied without uninstalling Thunder and discarding its data — and, because pointing Thunder at a fresh database yields one with no OAuth clients in it, nothing on the platform would be able to authenticate.

Verifying the rest of the guide against the charts turned up several more instructions that could not work as written (detailed under Approach).

Review of this PR also surfaced a gap the documentation could only describe, not fix: the Agent Manager built its database connection string with no SSL mode, so pgx applied its libpq default of `prefer` — TLS when the server offers it, and a **silent** fall back to plaintext when it does not. No chart value could require it. Rather than ship a guide that documents that as unavoidable, this PR closes it.

Related https://github.com/wso2/agent-manager/issues/1058

## Goals

- Make the guide produce a production deployment when followed top to bottom, rather than requiring a second pass.
- Put every production setting at the step that applies it, so a setting is never described after the point where it had to be chosen.
- State plainly which decisions are irreversible, and front-load them.
- Correct the instructions that were factually wrong about the charts.
- Let an operator require TLS on the database connection, so it fails closed instead of downgrading silently.

## Approach

**Production configuration moved inline.** Each step that has a development/production split now offers both as tabs. The tab groups share a `groupId`, so choosing "Production" once carries through the page; the same mechanism syncs the Let's Encrypt/Corporate CA choice between the TLS issuer step and the per-environment Thunder step, which previously relied on the reader remembering their earlier choice.

**Two new planning subsections.** *Decisions you cannot change later* is a table of the five settings frozen at first boot, each linking to the step that fixes it. *Platform secrets* generates every internal credential up front with `openssl rand` and tabulates which chart sets each one and which other charts must match, so the steps reference variables instead of shipping well-known placeholder secrets.

**Thunder and OpenBao gained real production paths.** Thunder can now be installed against external PostgreSQL for all three of its logical databases, with the password supplied by `passwordRef` so it stays out of Helm release history. OpenBao's production tab covers the sealed install, `bao operator init`/unseal, and the KV mount, Kubernetes auth role, policies, and secret seeding that dev mode had been performing invisibly — a sealed installation does not create the `secret/` mount on its own, which the guide never mentioned.

**The trailing section became a reference.** It now leads with a checklist recording where each item is applied and whether it can still be applied to a running platform, followed only by concerns that belong to no single step: identity, gateway hardening, build workflows, high availability, cluster security, and agent isolation.

**Whole-page development/production tabs were considered and rejected.** Docusaurus emits headings from *every* tab into the table of contents regardless of which tab is selected (verified by building a test page and inspecting the output), so that approach would have listed every install step twice and left anchor links resolving into tabs that are not visible. It would also have duplicated content that the Quick Start, k3d, and VM guides already cover.

**Requiring TLS on the database connection (chart + service).** `DB_SSL_MODE` and `DB_SSL_ROOT_CERT` are threaded through the config loader and `makeConnString`, surfaced as `postgresql.external.sslMode` and `.sslRootCert`, and rendered for **both** workloads that open a connection — the API deployment and the migration job, which runs the same binary through the same DSN builder.

Both default to empty, which emits no query string at all: the DSN stays byte-identical to earlier builds, pgx keeps its `prefer` default, and `PGSSLMODE` still works. A full `helm template` diff against `main` for the default and external-without-`sslMode` cases is identical except the chart's own random secrets. The values are also ignored while `postgresql.enabled` is true, since the bundled PostgreSQL listens without TLS and `require` would otherwise fail the API pod and the post-install migration hook on a default install. An invalid mode is rejected at config load with the accepted values listed, rather than surfacing as an opaque driver error on first connect.

`verify-ca` and `verify-full` work as well: pgx falls back to the image trust store, so `verify-full` needs no mounted CA for a database whose certificate chains to a public one — `sslRootCert` accepts the literal `system` for that, or a path to a PEM mounted through the chart's existing `volumes`/`volumeMounts` values. Those two values were rendered only by the API deployment, so the migration job now renders them too; without that, a private CA reached the API but not the post-install migration hook, failing the install with `x509: unknown authority`. A dedicated, purpose-named CA value is still deferred as unnecessary now that the generic hook covers both workloads.

Note the value is `sslMode` (camelCase, matching every other key in this chart) where the `thunderid` subchart uses `sslmode`. That subchart's default of `require` is not usable precedent for defaulting the same way here: this platform forces all three Thunder databases to SQLite, so that block is dead config.

**Corrections made while verifying against the charts:**

| Issue | Correction |
|---|---|
| OpenChoreo API was published on the control-plane gateway | Disabled via `openchoreoApi.http.enabled=false`; the API is reached over cluster DNS, and its hostname, DNS record, and access-table row are removed |
| OpenSearch persistence values used the `opensearch` alias | The subchart alias is `openSearch`; the documented values were silently ignored. The guide also described the storage as ephemeral, when the chart renders a small persistent volume |
| Changing the OpenSearch password appeared to be a single change | The logs module also passes the password to OpenSearch through `openSearch.extraEnvs`; without a matching override the two sides disagree and all observability queries fail authentication |
| Per-environment Thunder hostnames had no TLS or DNS coverage | Added `*.thunder.<base>` to the control-plane gateway certificate and the DNS records, with a note that RFC 4592 makes an explicit `thunder.<base>` record mandatory once that wildcard exists |
| Gateway encryption keys were documented as a chart value | The extension does not expose one; the guide now describes the values-ConfigMap procedure and that it must be re-applied after chart upgrades |
| Evaluation publishing was described as a static API key | It authenticates with the `amp-publisher-client` OAuth2 credentials read from OpenBao at workflow runtime |
| `kubectl wait` on the PostgreSQL StatefulSet was unconditional | No StatefulSet exists with an external database; the wait is now part of the development variant only |
| The guide said TLS could not be required from the client | It can now; the production path sets `sslMode=require` and the warning explains what `require` does and does not verify |

The shared Phase 2 partial (`_amp-installation.mdx`) is used by the k3d guide as well, so its production content is behind a `production` prop. The k3d guide is unchanged in rendered output.

## User stories

- As a platform engineer deploying Agent Manager to a production cluster, I can follow the guide once and end up with a production-grade deployment, instead of installing and then discovering that some hardening steps required a reinstall.
- As a platform engineer, I can see before I start which decisions are permanent, so I can gather the databases and credentials I need in advance.
- As an evaluator, I can still follow the same page with development-grade defaults by choosing the evaluation tabs.
- As a platform engineer, I can require TLS on the database connection so a server that stops enforcing it causes a visible failure rather than a silent downgrade to plaintext.

## Release note

Rewrote the On Your Environment installation guide as a production deployment guide, with production configuration presented inline at each step, an explicit list of settings that cannot be changed after installation, and generated platform secrets replacing the shipped defaults.

Added `postgresql.external.sslMode` and `postgresql.external.sslRootCert` to the Agent Manager chart, allowing TLS to be required (and the server certificate verified) on the connection to an external PostgreSQL database. Defaults are unchanged: without these values the connection behaves exactly as before.

## Documentation

This PR is the documentation change. The affected page is `documentation/docs/getting-started/on-your-environment.mdx`, published at `/docs/next/getting-started/on-your-environment/`.

## Training

N/A — no training content covers platform installation.

## Certification

N/A — installation procedure changes are not covered by certification exams.

## Marketing

N/A — documentation correction and restructure, no new product capability to promote.

## Automation tests

- Unit tests
  > `config` and `db` gain coverage for the new settings — `config_loader_test.go` (+71) for the allowlist and whitespace handling, and a new `db_test.go` (~130) for DSN construction, in a package that previously had no tests. `make test-unit` passes: `ok config (77.7% coverage)`, `ok db (14.3%)`. `make codegenfmt-check` is clean, and `go mod tidy` produces no drift.
- Integration tests
  > The chart's render assertions (`deployments/helm-charts/wso2-agent-manager/tests/render.sh`, run by CI) gain 9 cases covering the backward-compatibility contract: no TLS env rendered by default or for an external database without `sslMode`; `sslMode` and `sslRootCert` reaching **both** the API deployment and the migration job; and `sslMode` being ignored for the in-cluster database. All 19 assertions pass, including the pre-existing ones.
  >
  > The service was additionally exercised against real PostgreSQL containers on the migration-job code path (`-migrate -server=false`): plaintext with no settings runs all 36 migrations; plaintext with `require` fails with `server refused TLS connection`; an invalid mode is rejected before connecting; a TLS server with `require` connects; `verify-full` without a CA fails with `x509: unknown authority` and succeeds once `sslRootCert` is supplied.
  >
  > For the documentation, the Docusaurus production build runs with `onBrokenLinks: 'throw'`, validating every internal link and anchor. Rendered output was asserted for both affected pages to confirm production content appears only on the production guide and the k3d guide keeps its development variants.

## Security checks

- Ran FindSecurityBugs plugin and verified report? N/A — Java tooling; this is Go and Helm. The Go changes are covered by the repository's `golangci-lint` configuration in CI
- Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes — the guide generates credentials with `openssl rand` at install time and references them through shell variables and Kubernetes Secret references; no real credential values are committed. The database password continues to be read from a Kubernetes Secret, and the new TLS settings carry no secret material
- Note on the security posture of the change: `require` encrypts but does not authenticate the server, so `verify-full` with `sslRootCert` is documented as the stronger option, and server-side enforcement is still recommended alongside the client setting because the client cannot downgrade it

## Samples

N/A — no samples relate to this change.

## Related PRs

None.

## Migrations (if applicable)

No migration is required for existing installations, and no action is forced by the chart change: with `sslMode` unset the rendered manifests and the connection string are unchanged, which is asserted by both a full `helm template` diff against `main` and the chart render tests.

Operators who want to adopt it should confirm their database presents a TLS certificate before setting `require`, since the connection will then refuse to fall back. The guidance otherwise targets new deployments; existing installations remain supported, though operators may want to review the Production Reference checklist — in particular the OpenSearch admin password and the per-environment Thunder DNS and certificate coverage.

## Test environment

macOS. Documentation build via the repository's Docusaurus toolchain (`npm run build` in `documentation/`). Go build and unit tests via `make test-unit`; formatting and codegen via `make codegenfmt-check`. Chart rendering with Helm (local v4.0.5 — CI pins 3.19.4; only long-standing `with`/`trim`/`nindent`/`include` constructs are used) plus `helm lint --strict`. The TLS behaviour was exercised against PostgreSQL 16 containers, with and without a server certificate.

Not verified: a real RDS, Cloud SQL, or Azure Flexible Server instance; `golangci-lint` with the repository configuration (the local binary predates the config's schema version); and `verify-ca` separately from `verify-full`, which shares the same driver branch.

The install procedures themselves have not yet been executed end to end on a cloud cluster — the newly written Thunder external-PostgreSQL and sealed-OpenBao paths in particular have not been run, and remain the highest priority for a validation run before this guidance is relied upon.

## Learning

The structural question — inline production configuration versus development/production tabs spanning the page — was settled by looking at how comparable self-hosted products document the same split, and by testing the Docusaurus behaviour directly.

[Keycloak](https://www.keycloak.org/server/configuration-production) is the closest analogue, an identity product whose development mode is backed by an embedded database; it keeps a standalone production configuration guide and links to it from the getting-started pages rather than tabbing them together. [Temporal](https://docs.temporal.io/self-hosted-guide/deployment) organises by deployment method and embeds production guidance inline within each. [OpenChoreo](https://openchoreo.dev/docs/getting-started/try-it-out/on-your-environment/#production-configuration), which this platform builds on, keeps a lean install flow with a production section that links out per component. None of them tab an entire install page.

On the database TLS change, the alternative considered was a generic `extraEnv` hook to inject `PGSSLMODE`, which needs no Go change and was verified to work. It was rejected because the migration job has no `extraEnv` hook either, so both designs require editing `db-migration-job.yaml` — at which point the typed value costs about fourteen further lines of Go and buys load-time validation, a documented `values.yaml` entry, and parity with the `thunderid` subchart. Confirming that the migration job shares `makeConnString` (it runs the same binary with `-migrate`, over `gormigrate`, with no second DSN builder anywhere in the repository) was what made the change small enough to be worth doing properly.

The deciding technical detail was Docusaurus' [table of contents behaviour](https://docusaurus.io/docs/markdown-features/tabs): headings inside `<TabItem>` are collected into the page TOC regardless of the selected tab. A scratch page confirmed both variants of a heading appear, which rules out the whole-page tab approach for a guide whose structure is its ten numbered steps. Tabs are used only where a genuine either/or lives inside a single step, which is what they are designed for.

## Summary by CodeRabbit

- **New Features**
  - Added production-first installation guidance with separate Production and Evaluation paths across setup steps.
  - Enabled secure external PostgreSQL connections with configurable TLS settings, including CA certificate support, for runtime services and database migrations.
- **Documentation**
  - Expanded environment planning, TLS and issuer guidance, Thunder/PostgreSQL/Evaluation instructions, registry requirements, and troubleshooting.
  - Added warnings about settings requiring reinstallation and clarified secret, certificate, and OpenBao persistence expectations.
- **Bug Fixes**
  - Added validation for unsupported PostgreSQL TLS modes to prevent invalid configurations.

### DIFF
--- a/agent-manager-service/.env.example
+++ b/agent-manager-service/.env.example
@@ -25,6 +25,15 @@ DB_USER=agentmanager
 DB_PASSWORD=agentmanager
 DB_NAME=agentmanager
 
+# Database TLS Configuration (Optional)
+# DB_SSL_MODE: disable|allow|prefer|require|verify-ca|verify-full.
+# Unset leaves the driver default ("prefer"), which negotiates TLS when the
+# server offers it and silently falls back to plaintext when it does not.
+# DB_SSL_MODE=
+# DB_SSL_ROOT_CERT: PEM CA bundle path for verify-ca/verify-full, or "system"
+# to validate against the container trust store.
+# DB_SSL_ROOT_CERT=
+
 # Database Pool Configuration (Optional)
 # GORM_SKIP_DEFAULT_TRANSACTION=true
 # GORM_SLOW_THRESHOLD_MILLISECONDS=200
--- a/agent-manager-service/config/config.go
+++ b/agent-manager-service/config/config.go
@@ -245,6 +245,15 @@ type POSTGRESQL struct {
 	User     string
 	DBName   string
 	Password string `json:"-"`
+	// SSLMode is the libpq/pgx "sslmode" connection parameter
+	// (disable|allow|prefer|require|verify-ca|verify-full). Empty means the
+	// parameter is omitted from the connection string, which leaves the driver
+	// default ("prefer") — and any PGSSLMODE in the environment — in effect.
+	SSLMode string
+	// SSLRootCert is the libpq/pgx "sslrootcert" connection parameter: a path to
+	// a PEM CA bundle, or the literal "system" to use the image trust store.
+	// Empty omits the parameter, so verification falls back to the system pool.
+	SSLRootCert string
 	DbConfigs
 }
 
--- a/agent-manager-service/config/config_loader.go
+++ b/agent-manager-service/config/config_loader.go
@@ -21,6 +21,7 @@ import (
 	"log/slog"
 	"net/url"
 	"os"
+	"strings"
 
 	"github.com/joho/godotenv"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -79,6 +80,11 @@ func loadEnvs() {
 		User:     r.readRequiredString("DB_USER"),
 		Password: r.readRequiredString("DB_PASSWORD"),
 		DBName:   r.readRequiredString("DB_NAME"),
+		// Left empty by default so the connection string stays byte-identical to
+		// pre-DB_SSL_MODE builds: pgx applies its libpq-compatible "prefer"
+		// default and still honours PGSSLMODE/PGSSLROOTCERT from the environment.
+		SSLMode:     strings.TrimSpace(r.readOptionalString("DB_SSL_MODE", "")),
+		SSLRootCert: strings.TrimSpace(r.readOptionalString("DB_SSL_ROOT_CERT", "")),
 	}
 	config.POSTGRESQL.DbConfigs = DbConfigs{
 		// gorm configs
@@ -265,11 +271,39 @@ func loadEnvs() {
 	validateInstrumentationURL(config, r)
 	validateObserverURLs(config, r)
 	validateResourceLimitsConfig(config, r)
+	validatePostgresTLSConfig(config, r)
 	validateAgentWorkloadCORSConfig(agentWorkloadConfig, r)
 
 	r.logAndExitIfErrorsFound()
 
 	slog.Info("configReader: configs loaded")
+}
+
+// validPostgresSSLModes is the set of libpq sslmode values pgx accepts. Kept as
+// an explicit allowlist so a typo fails config load with a clear message instead
+// of surfacing as an opaque driver parse error at first connect.
+var validPostgresSSLModes = map[string]struct{}{
+	"disable":     {},
+	"allow":       {},
+	"prefer":      {},
+	"require":     {},
+	"verify-ca":   {},
+	"verify-full": {},
+}
+
+func validatePostgresTLSConfig(cfg *Config, r *configReader) {
+	// loadEnvs already trims, so in the real flow this value is exactly what
+	// makeConnString puts in the DSN. Trimmed again here so the check holds for
+	// callers that build a Config directly rather than from the environment.
+	mode := strings.TrimSpace(cfg.POSTGRESQL.SSLMode)
+	if mode == "" {
+		return
+	}
+	if _, ok := validPostgresSSLModes[mode]; !ok {
+		r.errors = append(r.errors, fmt.Errorf(
+			"DB_SSL_MODE %q is not a valid PostgreSQL sslmode (disable, allow, prefer, require, verify-ca, verify-full)", mode,
+		))
+	}
 }
 
 func validateHTTPServerConfigs(cfg *Config, r *configReader) {
--- a/agent-manager-service/config/config_loader_test.go
+++ b/agent-manager-service/config/config_loader_test.go
@@ -246,6 +246,77 @@ func TestValidateObserverURLs(t *testing.T) {
 	}
 }
 
+func TestValidatePostgresTLSConfig(t *testing.T) {
+	tests := []struct {
+		name        string
+		sslMode     string
+		wantErrors  int
+		errContains string
+	}{
+		{
+			name:       "empty is allowed and leaves the driver default in place",
+			sslMode:    "",
+			wantErrors: 0,
+		},
+		{
+			name:       "whitespace-only is treated as unset",
+			sslMode:    "   ",
+			wantErrors: 0,
+		},
+		{
+			name:       "require accepted",
+			sslMode:    "require",
+			wantErrors: 0,
+		},
+		{
+			name:       "disable accepted",
+			sslMode:    "disable",
+			wantErrors: 0,
+		},
+		{
+			name:       "verify-full accepted",
+			sslMode:    "verify-full",
+			wantErrors: 0,
+		},
+		{
+			name:        "typo rejected",
+			sslMode:     "requrie",
+			wantErrors:  1,
+			errContains: "DB_SSL_MODE",
+		},
+		{
+			name:        "uppercase rejected because libpq is case sensitive",
+			sslMode:     "REQUIRE",
+			wantErrors:  1,
+			errContains: "is not a valid PostgreSQL sslmode",
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			cfg := &Config{POSTGRESQL: POSTGRESQL{SSLMode: tc.sslMode}}
+			r := &configReader{}
+			validatePostgresTLSConfig(cfg, r)
+
+			if len(r.errors) != tc.wantErrors {
+				t.Fatalf("expected %d errors, got %d: %v", tc.wantErrors, len(r.errors), r.errors)
+			}
+			if tc.errContains != "" {
+				found := false
+				for _, e := range r.errors {
+					if strings.Contains(e.Error(), tc.errContains) {
+						found = true
+						break
+					}
+				}
+				if !found {
+					t.Errorf("expected an error containing %q, got %v", tc.errContains, r.errors)
+				}
+			}
+		})
+	}
+}
+
 func TestLoadEnvs_ObserverConfig(t *testing.T) {
 	requiredEnv := map[string]string{
 		"OPEN_CHOREO_BASE_URL": "http://localhost/api/v1",
--- a/agent-manager-service/db/db.go
+++ b/agent-manager-service/db/db.go
@@ -132,6 +132,15 @@ func GetDB() *gorm.DB {
 
 func makeConnString(p config.POSTGRESQL) string {
 	params := url.Values{}
+	// Only set the TLS parameters when configured. Omitting them keeps the DSN
+	// byte-identical to earlier builds, so pgx's libpq-compatible "prefer"
+	// default (and any PGSSLMODE/PGSSLROOTCERT in the environment) still apply.
+	if p.SSLMode != "" {
+		params.Set("sslmode", p.SSLMode)
+	}
+	if p.SSLRootCert != "" {
+		params.Set("sslrootcert", p.SSLRootCert)
+	}
 	conn := &url.URL{
 		Scheme:   "postgres",
 		User:     url.UserPassword(p.User, p.Password),
--- a/agent-manager-service/db/db_test.go
+++ b/agent-manager-service/db/db_test.go
@@ -1,0 +1,130 @@
+// Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+//
+// WSO2 LLC. licenses this file to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package db
+
+import (
+	"testing"
+
+	"github.com/jackc/pgx/v5"
+
+	"github.com/wso2/agent-manager/agent-manager-service/config"
+)
+
+func basePostgresConfig() config.POSTGRESQL {
+	return config.POSTGRESQL{
+		Host:     "db.example.com",
+		Port:     5432,
+		User:     "agentmanager",
+		Password: "s3cr#t/pass",
+		DBName:   "agentmanager",
+	}
+}
+
+func TestMakeConnString(t *testing.T) {
+	tests := []struct {
+		name        string
+		sslMode     string
+		sslRootCert string
+		want        string
+	}{
+		{
+			name: "no TLS settings keeps the pre-existing DSN with no query string",
+			want: "postgres://agentmanager:s3cr%23t%2Fpass@db.example.com:5432/agentmanager",
+		},
+		{
+			name:    "sslmode is appended when set",
+			sslMode: "require",
+			want:    "postgres://agentmanager:s3cr%23t%2Fpass@db.example.com:5432/agentmanager?sslmode=require",
+		},
+		{
+			name:        "sslrootcert is appended alongside the mode",
+			sslMode:     "verify-full",
+			sslRootCert: "/etc/amp/db-ca/ca.crt",
+			want:        "postgres://agentmanager:s3cr%23t%2Fpass@db.example.com:5432/agentmanager?sslmode=verify-full&sslrootcert=%2Fetc%2Famp%2Fdb-ca%2Fca.crt",
+		},
+		{
+			name:        "sslrootcert alone is appended",
+			sslRootCert: "system",
+			want:        "postgres://agentmanager:s3cr%23t%2Fpass@db.example.com:5432/agentmanager?sslrootcert=system",
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			cfg := basePostgresConfig()
+			cfg.SSLMode = tc.sslMode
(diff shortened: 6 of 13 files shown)


RAVEENSR:
Followed up on the database TLS review comment by fixing it here rather than deferring to a separate PR, since the docs otherwise ship a statement that is about to become false. Pushed in fe2c59f2; title and description updated for the widened scope.

**What landed**

`DB_SSL_MODE` and `DB_SSL_ROOT_CERT` threaded through the config loader and `makeConnString`, surfaced as `postgresql.external.sslMode` / `.sslRootCert`, and rendered for **both** workloads that open a connection — the API deployment and the migration job. The migration job matters here: it runs the same binary with `-migrate` through the same DSN builder, so a mode set only on the API would have left the post-install hook connecting in plaintext.

**Backward compatibility is the load-bearing property.** Both values default to empty, which emits no query string at all — the DSN stays byte-identical, pgx keeps its `prefer` default, and `PGSSLMODE` still works. A full `helm template` diff against `main` for the default and external-without-`sslMode` cases is identical except the chart's own random secrets. The values are additionally ignored while `postgresql.enabled` is true, because the bundled PostgreSQL listens without TLS and `require` would otherwise fail the API pod and the migration hook on a default install.

`verify-ca` / `verify-full` are supported now rather than deferred: pgx falls back to the image trust store, so `verify-full` needs no mounted CA for a publicly-chained managed database — `sslRootCert` takes the literal `system` for that. Only a first-class CA-mounting value is deferred, because the migration job has no volume hook at all.

**On the earlier suggestion**

The original comment proposed setting an existing chart `sslmode` value; there wasn't one, which is why the first pass documented server-side enforcement instead. The alternative of a generic `extraEnv` to inject `PGSSLMODE` does work, but the migration job has no `extraEnv` hook either, so both designs require touching `db-migration-job.yaml` — at which point the typed value costs ~14 more lines of Go and buys load-time validation, a documented `values.yaml` entry, and parity with the `thunderid` subchart.

Two deliberate deviations worth a reviewer's eye: the key is `sslMode` (camelCase, matching every other key in this chart) where `thunderid` uses `sslmode`; and that subchart's `require` default is **not** precedent for defaulting the same way here, since this platform forces all three Thunder databases to SQLite, making that block dead config.

**Verification**

`make test-unit` passes (`config` 77.7%, `db` 14.3% — `db` had no tests before); `make codegenfmt-check` clean with no `go mod tidy` drift; `helm lint --strict` clean. I extended the chart render assertions with 9 cases pinning the compatibility contract — no TLS env by default, both values reaching both workloads when set, and `sslMode` ignored for the in-cluster database — and all 19 assertions pass including the pre-existing ones.

Also exercised against real PostgreSQL 16 containers on the migration path: plaintext with no settings runs all 36 migrations; plaintext with `require` fails with `server refused TLS connection`; an invalid mode is rejected before connecting; TLS with `require` connects; `verify-full` without a CA fails `x509: unknown authority` and succeeds once `sslRootCert` is set.

Not verified: a real RDS/Cloud SQL/Azure instance, `golangci-lint` with the repo config (local binary predates the config schema version), and `verify-ca` separately from `verify-full`, which shares the same driver branch.

One unrelated observation, not addressed here: `initDbConn` passes a `connpool` retry wrapper as `gorm.Config.ConnPool`, but the postgres driver's `Initialize` replaces it with its own `stdlib.OpenDB` pool because `Dialector.Conn` is nil, so that wrapper looks dead. Harmless for this change — the DSN is parsed from the same string either way — but probably worth its own issue.

RAVEENSR:
Rebased onto `main` again to pick up #1443, which documented the JWT dev-mode auth bypass. It added its item to the *Production Considerations* list that this PR deletes, so resolving in favour of either side alone would have dropped a security control — the same failure mode as #1435, and worth flagging for anyone else rebasing onto this branch.

The guidance is carried over, in two places: a row in the Production Reference checklist, and a *Confirm JWT signature verification is active* subsection under Identity and Access with the commands to check it.

**I did not copy the wording, because it does not match the code.** Verified against `agent-manager-service` and both charts:

- `config.IsLocalDevEnv = r.readOptionalBool("IS_LOCAL_DEV_ENV", false)` — the default is already `false`, so "Set `IS_LOCAL_DEV_ENV=false`" implies a default that does not exist.
- `middleware/jwtassertion/auth.go` takes the unverified path only when the JWKS URL is empty **and** `IsLocalDevEnv` is true. With neither set it returns `KEY_MANAGER_JWKS_URL must be configured for JWT validation` — so "when unset, JWT signature verification is skipped" is the opposite of what happens. It fails closed.
- The chart never sets `IS_LOCAL_DEV_ENV` on the API deployment, and `keyManager.jwksUrl` has a non-empty default which this guide overrides explicitly anyway.
- The observer is guarded at render time: `amObserver.auth.isLocalDevEnv` defaults to `false`, and while false the chart `fail`s unless `auth.jwksUrl`, `auth.issuer`, and `auth.audience` are all set.

So both services are secure by default and the bypass is opt-in, requiring two conditions. The subsection says that, and frames the task as confirming the bypass is off rather than turning it off. It also notes that `amp-db-migration` sets `IS_LOCAL_DEV_ENV=true` — harmless, since that job runs `-migrate -server=false` and validates no tokens; someone auditing env vars will otherwise find it and reasonably worry.

Two follow-ups for @AnoshanJ, not changed here since they are outside this PR's scope: the same wording is now in `on-k3d.mdx`, and it names raw environment variables rather than the Helm values a reader of either guide would actually set (`agentManagerService.config.keyManager.jwksUrl`, `amObserver.auth.isLocalDevEnv`). Happy to send a small follow-up PR correcting both if you would like.

**Verification after the rebase:** docs build passes with `onBrokenLinks: 'throw'`; all 19 chart render assertions pass; `make test-unit` passes. I asserted the rendered HTML of both pages to confirm #1435's MCP flags and observer issuer, #1443's JWT guidance, and this PR's `sslMode=require` all survive, that the superseded section is gone, and that `on-k3d` still shows only development variants.

RAVEENSR:
Thanks — the first of these was a genuine bug in this PR, now fixed in 2c8b8a21. The second is not applicable as written.

**1. CA volume missing from the migration job — correct, and worse than a nitpick.**

Verified: `templates/agent-manager-service/deployment.yaml` renders `.Values.volumeMounts` and `.Values.volumes`, and `templates/jobs/db-migration-job.yaml` rendered neither. Since both workloads open a connection to the same database, `verify-ca`/`verify-full` against a **private** CA was broken end to end — the API could read the mounted PEM, the migration job could not, and because it runs as a post-install hook the install failed with `x509: unknown authority` before anything came up. The documentation in this PR compounded it by telling operators to mount the PEM through exactly those values.

The job now renders both, emitting nothing when they are empty so the default install is unchanged. Three render assertions were added — a CA volume reaching the API, the same volume reaching the migration job, and no volume being invented for the job by default — because this only breaks with a private CA and would otherwise regress unnoticed. 22 assertions pass, `helm lint --strict` clean.

This also removes the reason I had given for deferring a purpose-named CA value; the PR description is updated. A dedicated value now looks unnecessary rather than blocked, since the generic hook covers both workloads.

**2. `existingSecret` for `oidc.clientSecret` and `amObserver.observer.idpClientSecret` — no such option exists.**

Checked both charts. `agentManagerService.config.oidc` exposes only `tokenUrl`, `clientId`, and `clientSecret`; `amObserver.observer` exposes only `idpClientSecret`. Neither has an `existingSecret` counterpart, so there is nothing to document. The two places the guide does mention it — `agentManagerService.config.thunder.existingSecret` and the gateway extension's `agentManager.idp.existingSecret` — are precisely the ones where the chart supports it.

The underlying point is fair, though: secret handling is inconsistent across these charts, and two of the four platform client secrets can only be passed inline, landing them in Helm release history. That is a chart change rather than a documentation one, so I have left it out of this PR and will raise it separately alongside the other chart gaps found here.

RAVEENSR:
Rebased onto `main` for #1439, which makes part of this guide wrong and part of it possible. Updated in c1aa46c6.

**The stale claim.** This guide carried a note saying a split ingress/egress topology was "planned". #1439 shipped it, so that note is replaced with what now exists: the three roles (`INGRESS`, `EGRESS`, `BOTH`), how to create a split environment, the cap of one ingress-capable gateway per environment, and the deprecated `REGULAR`/`AI` aliases. It links to [Gateway topology](../administration/environment-management.mdx#gateway-topology) rather than restating the per-environment procedure documented there.

**A new irreversible decision.** Verified in `gateway-bootstrap-job.yaml`: on an existing gateway the job performs a read-only drift check and logs a warning, and the role is written only at registration. So an environment's topology cannot be switched in place — moving to split means deleting and recreating the environment. That is exactly the class of thing this PR's *Decisions you cannot change later* table exists for, so it is now a row there.

**The role is now set explicitly on the production install** (`--set gateway.type=BOTH`) rather than left to the default. It matches the chart default and changes nothing today, but the value is immutable once registered and its default moved from `REGULAR` to `BOTH` in this very release — pinning an immutable value seems right for a production path.

**A rollback hazard worth surfacing.** The bootstrap job's hook is now `pre-install,pre-upgrade,pre-rollback`, so `helm rollback` runs it and can fail, leaving the release in `pending-rollback`, when the Agent Manager API is unreachable — which is plausibly the situation prompting the rollback. Documented with the recovery (`helm rollback` again once the API is back, or `--no-hooks`).

**One code conflict, resolved deliberately.** #1439 deleted `validateGatewayRuntimeConfig` from `config_loader.go` along with the `gatewayRuntime` values block it validated, while this PR's TLS commit had added `validatePostgresTLSConfig` adjacent to it. Kept the deletion and my addition; the `strings` import went with their deletion and had to come back for mine. Rebuilt and re-ran the suites rather than assuming the resolution was right.

I deliberately did **not** document the split procedure itself here — #1439 already covers it in `environment-management.mdx`, and this page is the single-cluster install, not per-environment administration. Its note that "operator docs for the split topology should follow once the topology is exercised beyond e2e" still stands for anything deeper than the pointer added here.

**Verification:** docs build passes with `onBrokenLinks: 'throw'`; `make test-unit` passes (`config` 76.8%, `db` 14.3%); 22/22 chart render assertions pass. I asserted the rendered HTML on both pages — the production page has the roles, split pointer, ingress cap, immutability, rollback warning and pinned `gateway.type`, the removed claim is gone, and `on-k3d` picks up none of it and keeps its development variants.

jhivandb (approved):

Issue #1433, opened 2026-07-28
[Improvement]: Remove the Kubernetes ServiceAccount token from the evaluation-job's main container


AnoshanJ:
### Please select the area the issue is related to

Area/Evaluation (Evaluation jobs, evaluators), Area/Infrastructure (Environments, deployment pipelines, infra resources, gateway integration)

### Please select the aspect the issue is related to

Aspect/Evaluation (Evaluators and evaluation jobs), Aspect/Infra Resources (Infrastructure resources)

### Suggested Improvement

The evaluation-job pod's `main` container runs untrusted, `exec()`'d/`eval()`'d custom evaluator code and never calls the Kubernetes API, but it still gets a `workflow-sa` ServiceAccount token auto-mounted, since `automountServiceAccountToken: false` isn't set anywhere.

Fix: set `automountServiceAccountToken: false` pod-wide, and use Argo's `executor.serviceAccountName` so only the `wait` sidecar (which does need to report `WorkflowTaskResults`) gets a token — via a manually-provisioned bound-token Secret, since Kubernetes 1.24+ no longer auto-creates these for a ServiceAccount.

Resolved in #1447.

### Related Issues

#1427

Issue #1428, opened 2026-07-28
[Improvement]: Add basic content validation for custom evaluator source at creation time


AnoshanJ:
### Please select the area the issue is related to

Area/Evaluation (Evaluation jobs, evaluators)

### Please select the aspect the issue is related to

Aspect/Evaluation (Evaluators and evaluation jobs)

### Suggested Improvement

`POST /orgs/{orgName}/evaluators/custom` and the corresponding update endpoint currently validate only `displayName`/`type`/`level` and that `source` is non-empty (`agent-manager-service/utils/utils.go:1426-1449`). The content of `source` is never inspected — it's persisted verbatim and later `exec()`'d or `eval()`'d unmodified in `evaluation-job/main.py`.

Suggested improvement: add a basic content check (e.g. flagging high-risk patterns/imports such as `os`, `subprocess`, `socket`, `ctypes` via regex or similar) at creation/update time, to catch unintentional or low-effort misuse earlier and give clearer feedback than a runtime failure.

Note: pattern-based checks like this are easy to bypass (e.g. via `__import__`, `getattr`) and shouldn't be treated as a security boundary on their own — real containment needs to come from sandboxing/isolating the execution environment, not from filtering the source text. This is a defense-in-depth/UX improvement, not a fix for the underlying trust-boundary issue.

Issue #1427, opened 2026-07-28
[Improvement]: Restrict evaluation-job workflow pod egress with a NetworkPolicy


AnoshanJ:
### Please select the area the issue is related to

Area/Evaluation (Evaluation jobs, evaluators), Area/Infrastructure (Environments, deployment pipelines, infra resources, gateway integration)

### Please select the aspect the issue is related to

Aspect/Evaluation (Evaluators and evaluation jobs), Aspect/Infra Resources (Infrastructure resources)

### Suggested Improvement

The evaluation-job pod (which runs user-submitted custom evaluator code via `exec()`/`eval()`) currently has no network restrictions — no `NetworkPolicy` scopes its egress. It can reach any service on the cluster network, including the Kubernetes API server, the database, and internal services it has no functional need for.

Suggested fix: add a `NetworkPolicy` limiting this pod's egress to only what it actually needs:
- the traces-observer-service (`amp-observer`) to fetch trace data to evaluate
- the LLM gateway, for `llm_judge` evaluators
- agent-manager-service's publisher endpoint, to publish scores

Everything else (K8s API server, database, other internal services) should be denied. This also reduces the impact of any credential exposed to the pod (e.g. the Kubernetes ServiceAccount token), since a credential that can't reach its target service is far less useful even if it's readable.

PR #1447, opened 2026-07-30
Harden evaluation job security and custom evaluator validation

### LINKED ISSUE #1426: [Improvement]: Separate the scheduler and score-publishing OAuth2 credentials for monitor evaluation

Area/Evaluation (Evaluation jobs, evaluators), Area/Org & User Management (Projects, organizations, identities, RBAC, multi-tenancy)

Aspect/Evaluation (Evaluators and evaluation jobs), Aspect/Identity & Access (Identities, identity providers, roles, permissions, tenancy)

### Suggested Improvement

The per-org publisher OAuth2 credential (`amp-publisher-<ouID>`, provisioned in `agent-manager-service/services/publisher_credential_provisioner.go`) is currently reused for two different purposes:

1. The scheduler's own calls to OpenChoreo (`CreateWorkflowRun`, `GetWorkflowRun`), which need the `amp-monitor-scheduler` ClusterAuthzRole grant (`workflowrun:create`, `workflowrun:view`, `namespace:view`).
2. Score publishing from the evaluation-job pod back to agent-manager-service, which only needs to call the one `aud`-scoped publisher route.

Because both purposes share the same client_id/secret, the credential injected into the evaluation-job pod (which runs untrusted, user-submitted evaluator code) carries the scheduler's OpenChoreo-level grants too, even though score publishing never needs them. `namespace:view` in particular is a list operation, so this credential can enumerate namespaces beyond its own org.

Suggested fix: provision two separate credentials per org — a scheduler-only credential that never leaves agent-manager-service, bound to `amp-monitor-scheduler`; and a publisher-only credential, bound to no OpenChoreo role at all, that is the only one ever injected into the evaluation-job pod's environment.

### Related Issues

_No response_

AnoshanJ:
This PR hardens the evaluation job execution environment by reducing its privileges, restricting network access, and validating custom evaluator source before it is stored and executed.
These changes address findings from a security review of the monitor evaluation pipeline.

Problems:
- The evaluation-job pod's OAuth2 credential (`amp-publisher-<ouID>`) was the same credential the scheduler uses for its own OpenChoreo calls, so it carried the scheduler's `amp-monitor-scheduler` grants too — `workflowrun:create`, `workflowrun:view`, and `namespace:view` (a list operation) — even though publishing scores only ever needs to call one `aud`-scoped route.
- The evaluation-job pod's egress wasn't scoped to the handful of services it actually calls (the traces observer, Thunder, the publisher endpoint, the LLM gateway). It could reach anything else on the cluster network. However note, there is an egress rule for the Kubernetes API server, which the `wait` sidecar needs to report each run's status.
- Custom evaluator source is later `exec()`'d/`eval()`'d as-is by the evaluation job, but the create/update API only checked that `displayName`/`type`/`level` were present and `source` was non-empty — the content itself wasn't inspected before being persisted.
- The evaluation-job pod's `main` container — the one running the untrusted, `exec()`'d/`eval()`'d evaluator code — still got a Kubernetes ServiceAccount token auto-mounted, even though nothing in it calls the Kubernetes API.

Changes:
- Split the scheduler's OpenChoreo credential from the evaluation-job's publishing credential, so the evaluation-job pod's credential only carries the grants it actually needs.
- Add a NetworkPolicy restricting the evaluation-job pod's egress to the specific services it calls.
- Add a basic content check on custom evaluator source at creation/update time, rejecting high-risk imports with a clear error instead of a runtime failure.
- Remove the ServiceAccount token from the `main` container (`automountServiceAccountToken: false` pod-wide), while still giving the `wait` sidecar its own token via `executor.serviceAccountName` and a manually-provisioned bound-token Secret (Kubernetes 1.24+ no longer auto-creates these for a ServiceAccount).

Resolves #1426, #1427, #1428, #1433

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary by CodeRabbit

* **New Features**
  * Added separate scheduler credentials with automatic provisioning, rotation, and recovery.
  * Added application creation, deletion, and client-secret regeneration.
  * Added configurable NetworkPolicies for evaluation jobs.

* **Bug Fixes**
  * Improved fresh-install setup by creating required workflow namespaces automatically.
  * Improved evaluation workflow service-account handling.

* **Security**
  * Blocked risky imports in custom evaluator code.
  * Restricted evaluation-job network access to approved destinations.

* **Documentation**
  * Documented evaluator network restrictions, disallowed imports, and NetworkPolicy enforcement requirements.


deployments/helm-charts/wso2-amp-evaluation-extension/templates/evaluation-job-networkpolicy.yaml
```diff
@@ -0,0 +1,63 @@
...
+      app.kubernetes.io/name: amp-monitor
+      app.kubernetes.io/component: evaluation-job
+  policyTypes: [Egress]
+  egress:
+    # DNS resolution for all the service hostnames below.
+    - ports:
+        - port: 53
+          protocol: UDP
+        - port: 53
+          protocol: TCP
```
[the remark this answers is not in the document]
AnoshanJ: Left this as-is on purpose, matches the same unscoped DNS rule already used in agent-api.yaml, so it stays consistent with the existing pattern here.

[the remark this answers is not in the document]
AnoshanJ: Addressed

jhivandb (dismissed):

agent-manager-service/services/publisher_credential_provisioner.go
```diff
@@ -326,6 +329,102 @@ func (p *publisherCredentialProvisioner) provisionCredentials(ctx context.Contex
 	}, nil
 }
 
+// provisionSchedulerCredentials provisions the scheduler-only credential, bound to schedulerRoleName; never injected into the eval-job pod.
+func (p *publisherCredentialProvisioner) provisionSchedulerCredentials(ctx context.Context, ouID, orgUUID string) error {
+	existing, err := p.schedulerCredRepo.GetByOrgName(ouID)
```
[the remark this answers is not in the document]
AnoshanJ: This mirrors the exact same pattern followed repo wide. This would need a separate repo wide refactor. Leaving as-is for now.

deployments/helm-charts/wso2-amp-evaluation-extension/templates/evaluation-job-networkpolicy.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+    # LLM gateway for llm_judge evaluators — unconditional since NetworkPolicy can't scope by workflow.
+    - to:
+        - namespaceSelector:
+            matchLabels:
+              amp.wso2.com/api-platform-gateway: "true"
+      ports:
+        - port: {{ .Values.networkPolicy.evaluationJob.llmGateway.port }}
+          protocol: TCP
```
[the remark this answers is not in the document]
AnoshanJ: This namespace is dedicated per-environment gateway dataplane. Pinning to the gateway-runtime pod label specifically would just couple us to an internal implementation detail that could shift between OpenChoreo versions. Leaving as is for now

deployments/helm-charts/wso2-amp-evaluation-extension/templates/evaluation-job-networkpolicy.yaml
```diff
@@ -0,0 +1,87 @@
...
+    {{- $kubeSvc := lookup "v1" "Service" "default" "kubernetes" }}
+    {{- if $kubeSvc }}
+    # wait reports WorkflowTaskResults to the API server; no pod backs it, so only ipBlock matches.
+    - to:
+        - ipBlock:
+            cidr: {{ $kubeSvc.spec.clusterIP }}/32
+      ports:
+        - port: 443
+          protocol: TCP
+    {{- end }}
```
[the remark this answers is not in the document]
AnoshanJ: Fixed — added an explicit `networkPolicy.evaluationJob.apiServer.cidr` override that takes priority over the lookup, so RBAC-restricted installers (or plain `helm template`) can set it explicitly instead of silently losing the rule.

deployments/setup/setup-amp-extensions.sh
```diff
@@ -80,12 +80,36 @@ install_thunder_extension() {
...
+    local dev_egress_cidr network_policy_enabled
+    dev_egress_cidr=$(docker network inspect "k3d-${CLUSTER_NAME}" \
+        --format '{{ (index .IPAM.Config 0).Subnet }}' 2>/dev/null || echo "")
+    if [[ -z "$dev_egress_cidr" ]]; then
+        echo "⚠️  Could not determine k3d docker network subnet for k3d-${CLUSTER_NAME} — disabling"
+        echo "   the evaluation-job egress NetworkPolicy for this install so local eval runs keep working."
+        network_policy_enabled="false"
+    else
+        network_policy_enabled="true"
+    fi
```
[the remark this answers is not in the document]
AnoshanJ: This script only runs for the local dev installation, and an intentional fallback. The tradeoff was to keep local eval runs from silently breaking if we can't determine the docker network subnet, rather than hard-failing the whole setup over it.

deployments/helm-charts/wso2-amp-evaluation-extension/templates/workflow-templates/workflow-monitor-evaluation.yaml
```diff
@@ -180,6 +184,17 @@ spec:
...
+      # Name matches Argo's executor.serviceAccountName lookup convention: "<saName>.service-account-token".
+      template:
+        apiVersion: v1
+        kind: Secret
+        metadata:
+          name: workflow-sa.service-account-token
+          namespace: ${metadata.namespace}
+          annotations:
+            kubernetes.io/service-account.name: workflow-sa
+        type: kubernetes.io/service-account-token
```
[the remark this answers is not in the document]
AnoshanJ: The content is fully idempotent (identical manifest every time), so worst case a race just causes one harmless reconcile retry, not data loss. Leaving as is for now

agent-manager-service/services/publisher_credential_provisioner.go
```diff
@@ -197,22 +211,29 @@ func (p *publisherCredentialProvisioner) resolveSecretRef(ctx context.Context, o
...
 	})
 	if err != nil {
 		return nil, err
 	}
 	return result.(*PublisherCredentials), nil
 }
 
-// provisionCredentials performs the DB lookup and, if needed, the full Thunder provisioning flow.
-func (p *publisherCredentialProvisioner) provisionCredentials(ctx context.Context, ouID, orgUUID string) (*PublisherCredentials, error) {
+// provisionPublisherCredentials provisions the eval-job pod's credential — bound to no OpenChoreo role.
```
nadheesh, line 235: **Is the split retroactive for existing/cloud orgs?** This comment says the publisher credential is "bound to no OpenChoreo role", but that's only true going forward. We removed the `EnsureClusterRoleBinding(publisherClientID, schedulerRoleName)` calls, but nothing *unbinds* publisher apps that the old code already bound. Every org provisioned before this PR still has `amp-publisher-<ouID>` bound to `amp-monitor-scheduler`, and that same credential is still injected into the eval-job pod — so untrusted evaluator code in those orgs can still mint a token carrying the scheduler grants.

The new NetworkPolicy mitigates this **only where the CNI enforces it** (per our own docs, EKS/AKS/GKE defaults don't). So on the cloud we may need a one-time revocation pass: enumerate existing publisher clients and call `DeleteClusterRoleBinding` (already exposed on the OC client) for each. Can we confirm whether cloud orgs are affected and whether we need to revoke them all? At minimum we should note this as a known residual + follow-up, since "split the credential" reads as fully closing the issue when it only does so for newly-provisioned orgs.

nadheesh (commented):
Reviewed the security hardening. The layered approach (credential de-scoping + NetworkPolicy egress + source tripwire + SA-token removal) is the right design and test coverage is good. A few points raised inline — the main one is whether the credential split needs a retroactive cleanup for already-provisioned (cloud) orgs. Details in the inline comments.

deployments/helm-charts/wso2-amp-evaluation-extension/templates/evaluation-job-networkpolicy.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+    {{- $apiServerCIDR := .Values.networkPolicy.evaluationJob.apiServer.cidr }}
+    {{- if not $apiServerCIDR }}
+    {{- with lookup "v1" "Service" "default" "kubernetes" }}
+    {{- $apiServerCIDR = printf "%s/32" .spec.clusterIP }}
+    {{- end }}
+    {{- end }}
+    {{- if $apiServerCIDR }}
```
nadheesh, line 83: **Silent-failure risk.** When `apiServer.cidr` is empty (the default) the CIDR comes from the `lookup` above, which returns empty during `helm template`, any dry-run render, or when the release SA lacks RBAC to read the `kubernetes` Service. In that case `$apiServerCIDR` is empty and this whole rule is omitted — the pod-wide egress lockdown then black-holes the `wait` sidecar's API-server access, so evaluation workflows hang/fail with no obvious cause.

Suggest failing loudly instead of rendering a silently-broken policy, e.g. `{{ required "networkPolicy.evaluationJob.apiServer.cidr must be set when the lookup is unavailable" $apiServerCIDR }}` in the enabled+empty-lookup path.

agent-manager-service/utils/utils.go
```diff
@@ -1423,6 +1423,34 @@ func SanitizeString(s string) string {
 	}, strings.ToLower(s))
 }
 
+// riskyEvaluatorSourcePatterns are heuristics only, not a security boundary — evaluator
```
nadheesh, line 1426: Good that this is explicitly labelled a heuristic, not a boundary. Two things worth considering:

1. **Blocklist gaps that undercut the message.** We block `socket` but not the *easier* network paths — `urllib.request`, `http.client`, `requests`, `ftplib`, `smtplib` all pass. Filesystem access via the `open()` builtin needs no import at all. And the `^\s*` anchor is defeated by `x=1; import os`, `\`-continued imports, or `__builtins__['__imp'+'ort__']('os')`. If we're going to keep the list, adding `urllib`/`http`/`requests`/`ftplib` at least covers the network story the docs lean on; otherwise we should be careful the docs don't imply more than a tripwire.
2. **Error surfacing / UX** — I confirmed the backend returns a 400 with the descriptive message (`source contains a disallowed pattern (import os); ...`) and the console surfaces it via the error `Alert` in `EvaluatorForm.tsx`, so a violating submit *does* get a clear message. But that's only *after* submit — there's no proactive hint in the source editor. Should we add a short note near the source field in the create/edit UI ("os/subprocess/socket/ctypes/importlib and __import__() are not allowed") so users aren't surprised? Do we already have one? If not, worth a follow-up UI ticket.

deployments/helm-charts/wso2-amp-evaluation-extension/values.yaml
```diff
@@ -49,3 +49,46 @@ ampEvaluation:
...
+  evaluationJob:
+    # -- No-op on clusters whose CNI doesn't enforce NetworkPolicy (e.g. plain flannel).
+    enabled: true
+
+    # -- Empty derives "workflows-<ampEvaluation.workflowNamespace>" (where OpenChoreo runs this
+    # environment's Argo workflows), not this chart's own release namespace.
+    namespace: ""
+
+    # -- DNS target on port 53, scoped rather than left wide open since this pod runs untrusted code.
+    dns:
```
nadheesh, line 65: **DNS footgun.** Scoping DNS egress to `kube-system` / `k8s-app=kube-dns` breaks resolution on clusters running NodeLocal DNSCache or a CoreDNS with different labels/namespace — and when DNS fails, *every* egress fails, so all evaluations mysteriously fail. It's values-configurable, which is fine, but given the failure mode is so opaque, can we add a troubleshooting note (alongside the CNI-enforcement note) pointing at this as the first thing to check when evaluations hang with connection errors?

jhivandb (dismissed):

agent-manager-service/services/publisher_credential_provisioner.go
AnoshanJ: Confirmed, this only affects newly-provisioned orgs, existing ones keep the old binding, and there's no delete API to revoke it programmatically. We'll open an issue in the relevant repo to track/document this for existing orgs.

deployments/helm-charts/wso2-amp-evaluation-extension/templates/evaluation-job-networkpolicy.yaml
AnoshanJ: Fixed — swapped the silent omit for `required`, so it now fails the install loudly instead of quietly breaking the wait sidecar.

agent-manager-service/utils/utils.go
AnoshanJ: Added urllib/http/requests/ftplib/smtplib to the blocklist — cheap and worth closing. The deeper bypasses (line continuation, dynamic __import__ construction) are a real limit of a regex tripwire, not fixable without actual static analysis, which the comment already calls out.

deployments/helm-charts/wso2-amp-evaluation-extension/values.yaml
AnoshanJ: Added a troubleshooting note pointing at DNS scoping as the first thing to check when evaluations fail with connection errors.

nadheesh (approved):

PR #1465, opened 2026-08-03
Fix production install guide for cloud clusters

RAVEENSR:
## Purpose

A validation run of the production install guide on a multi-node managed Kubernetes cluster reached the agent lifecycle — build, deploy, invoke, traces, evaluation — for the first time, and surfaced seven defects in the guide.

Every one of them is silent. The platform installs, all pods report healthy, `helm list` shows every release deployed, and yet a user-visible capability does not work. Most trace back to a single-node local default that the guide inherits without overriding, so they cannot reproduce on a k3d or Rancher Desktop install.

Related https://github.com/wso2/agent-manager/issues/1058

## Goals

Make the guide produce a working platform on a real cloud cluster: traces reaching storage, agents exporting spans, a usable invoke URL in the console, a correctly advertised gateway virtual host, OAuth token validation on deployed agents, and resource-quota guidance that does not fail builds.

## Approach

Seven fixes, each with a warning admonition explaining the failure mode so the next reader recognises the symptom rather than rediscovering the cause.

**Traces never reach storage.** The OpenTelemetry collector manifest hardcodes the placeholder OpenSearch password. Applied verbatim beside the generated-password override in the same step, the collector authenticates with the placeholder against an OpenSearch holding the real one. Logs are unaffected — their adapter reads the credential from a Secret — so the only symptom is a permanently empty Traces view, with `failed to json unmarshal body: invalid character 'U'` (the start of `Unauthorized`) buried in collector logs. Now substituted as the manifest is applied.

**Deployed agents cannot export spans.** `AMP_OTEL_ENDPOINT` is derived from a convention assuming the gateway runtime lives in an `<org>-<env>` namespace, but Step 7 installs the extension into the data-plane namespace. The name does not resolve and every span batch fails inside the agent, which otherwise keeps serving traffic normally. Now set via `otelEndpointOverride`.

**The console publishes an invoke URL the browser refuses.** Two independent causes combine: `agentManagerService.config.tlsEnabled` defaults to `false` and selects the Environment's `http` variant, and the endpoint-wiring step assigned port `443` to that plaintext variant. The result is `http://<host>:443`, blocked as mixed content from the HTTPS console. Both corrected, with a note that the Environment binding wholly replaces the data plane's rather than merging.

**The gateway advertises a local virtual host.** `gateway.vhost`/`gateway.hostname` default to k3d addresses, and the bootstrap job writes the registration only at first install — later runs log `already exists` and exit without reconciling, so a `helm upgrade` cannot fix it. Now set on the initial install, with that constraint documented.

**Deployed agents cannot validate OAuth tokens.** The gateway extension is installed before the environment's Thunder exists, so its issuer is unknown at that point and nothing registers Thunder as a key manager or identity provider. API-key auth still works, so the platform looks complete while the console shows *No identity providers configured*. Adds a step after Thunder provisioning that wires both, including the detail that Helm's indexed-array `--set` replaces the whole list and the internal key manager must be re-stated. Also documents the `ReadWriteOnce` controller volume that makes the resulting rolling restart deadlock with `Multi-Attach error` on a multi-node cluster.

**The documented ResourceQuota has no effect, and breaks builds where it would.** It targets the workflow-plane namespace, which holds the Argo controller rather than build pods; builds run in the per-environment `workflows-<env>` namespace. Applied there, it rejects every build at admission with `must specify limits.cpu for: init,main,wait`, because build pods declare no requests or limits — terminally failed in seconds, not the queueing the guide promised. The section now names the right namespace, requires a `LimitRange` alongside the quota, drops the incorrect claim, and notes the namespace only exists after the environment's first workflow run.

## User stories

As an operator following the production install guide on a cloud cluster, I get a platform whose traces, agent invocation, gateway registration and build quotas all work as documented, without having to diagnose silent failures myself.

## Release note

Fixed seven issues in the production installation guide that produced a platform which installed cleanly but had broken trace ingestion, agent span export, console invoke URLs, gateway virtual host registration, OAuth token validation, and build resource quotas on multi-node cloud clusters.

## Documentation

This PR *is* the documentation change: `documentation/docs/getting-started/on-your-environment.mdx` and its shared Phase 2 partial `documentation/docs/getting-started/_partials/_amp-installation.mdx`.

## Training

N/A — no training content covers the production install guide.

## Certification

N/A — installation-guide corrections, no impact on certification exams.

## Marketing

N/A — a documentation defect fix, not a promotable feature.

## Automation tests

 - Unit tests
   > N/A — documentation only, no code paths changed.
 - Integration tests
   > Verified by `npm run build` in `documentation/` (the site sets `onBrokenLinks: 'throw'`, so every internal link and anchor is validated) plus assertions over the rendered HTML of `on-your-environment/index.html` and `on-k3d/index.html`: 11 expected strings present, 3 removed strings absent, and 4 production-only strings confirmed absent from the k3d page, since the Phase 2 partial is shared between the two.

Each individual fix was also applied to a live cluster during the validation run and confirmed to resolve its symptom — traces began flowing, the identity provider registered, invoke succeeded over HTTPS — and the quota behaviour was verified in both directions: unconstrained in the documented namespace, and terminally failing a real build in the correct one.

## Security checks

 - Ran FindSecurityBugs plugin and verified report? N/A — no code changes

One change is security-relevant in the other direction: the collector previously authenticated with a well-known placeholder password taken from a shipped manifest, and the guide now substitutes the operator's generated credential.

## Samples

N/A — no sample changes.

## Migrations (if applicable)

N/A for new installs. Existing installs that followed the previous guide can apply the trace-collector password substitution, the `tlsEnabled` and endpoint-port corrections, and the Thunder key-manager wiring as `helm upgrade`s. The gateway virtual host is the exception: it is written at first registration only, so correcting it requires deleting and re-registering the gateway.

## Test environment

Managed Kubernetes 1.35 (3 nodes, 4 vCPU / 16 GB, Container-Optimized OS, kernel 6.12), managed PostgreSQL 16 over a private address with TLS required, a managed container registry with authenticated push, Let's Encrypt production certificates via DNS-01, OpenChoreo planes 1.1.1, gateway operator 0.10.1 with gateway chart 1.2.0-beta, Thunder 0.45.0, and all Agent Manager charts at a single nightly version. Docs build verified on Node 20, macOS.

## Learning

The common thread is that a value which is correct for the single-cluster local layout becomes wrong on a cloud cluster, and the platform reports success either way. Five of the seven cannot reproduce on a single-node, single-replica, non-TLS install at all — one needs a multi-node cluster, one needs two API replicas, one needs HTTPS, one needs a real registry, and one needs an agent to actually be invoked.

Two process lessons worth recording. First, the repo's own install scripts already set most of these values correctly, with comments explaining why — the guide had drifted from them, so diffing the guide against `deployments/setup/setup-gateway.sh` and `deployments/scripts/add-environment.sh` is a cheap way to find this class of gap. Second, the mixed-content URL was invisible to command-line testing with hand-constructed URLs and obvious the moment the console's own Try It panel was used; at least one check should exercise the UI rather than reconstructing what it would do.

---

## Summary by CodeRabbit

* **Documentation**
  * Clarified production installation requirements, including domain, TLS, HTTPS, gateway, and telemetry settings.
  * Documented secure OpenSearch credential injection for telemetry collection.
  * Added troubleshooting guidance for OpenSearch authentication and trace exports.
  * Documented connecting Thunder to the API gateway for identity and key management while preserving API-key authentication.
  * Expanded workflow resource guidance with dedicated build nodes, quotas, limits, and default resource behavior.
  * Added warnings about incorrect settings, first registration, and recovery from partial configuration.

### DIFF
--- a/deployments/values/oc-collector-configmap.yaml
+++ b/deployments/values/oc-collector-configmap.yaml
@@ -28,9 +28,18 @@ data:
 
     extensions:
       basicauth/opensearch:
-        client_auth: 
-          username: admin
-          password: ThisIsTheOpenSearchPassword1
+        client_auth:
+          # Injected by the observability-tracing-opensearch chart's default
+          # opentelemetry-collector.extraEnvs, which source them from the
+          # opensearch-admin-credentials Secret — the same Secret the observer,
+          # the Fluent Bit adapter and openSearchSetup already read. Never
+          # inline the credentials here: a literal keeps working on a local
+          # install (where the seeded password happens to match the chart
+          # default) while silently failing anywhere the password was changed,
+          # and the collector then drops every trace with
+          # "failed to json unmarshal body: invalid character 'U'".
+          username: ${env:OPENSEARCH_USERNAME}
+          password: ${env:OPENSEARCH_PASSWORD}
       health_check:
         endpoint: ${env:MY_POD_IP}:13133
   
--- a/documentation/docs/getting-started/_partials/_amp-installation.mdx
+++ b/documentation/docs/getting-started/_partials/_amp-installation.mdx
@@ -23,7 +23,7 @@ The Agent Manager installs as a set of Helm charts on top of OpenChoreo. The com
 2. **Extensions :** Secret Management, Observability, Evaluation extensions and the API Platform Gateway Extension.
 
 :::info Prerequisites
-Thunder (identity provider) must be installed before proceeding — see the Thunder installation step in Phase 1. The variables `THUNDER_PUBLIC_URL`, `THUNDER_INTERNAL_URL`, `CONSOLE_PUBLIC_URL`, `API_PUBLIC_URL`, `OBS_API_PUBLIC_URL`, `CONSOLE_PUBLIC_HOST`, `API_PUBLIC_HOST`, `OBS_API_PUBLIC_HOST`, and `INSTRUMENTATION_URL` must be set from the Configuration Variables section. The `*_PUBLIC_HOST` variables carry the bare hostname (no scheme or port) and become the HTTPRoute hostnames on the plane gateways.
+Thunder (identity provider) must be installed before proceeding — see the Thunder installation step in Phase 1. The variables `THUNDER_PUBLIC_URL`, `THUNDER_INTERNAL_URL`, `CONSOLE_PUBLIC_URL`, `API_PUBLIC_URL`, `OBS_API_PUBLIC_URL`, `CONSOLE_PUBLIC_HOST`, `API_PUBLIC_HOST`, `OBS_API_PUBLIC_HOST`, and `INSTRUMENTATION_URL` must be set from the Configuration Variables section. The production steps below additionally use `AGENTS_DOMAIN` — the domain deployed agents are served on, `agents.<your base domain>` — to build the gateway's hostname and virtual host. The `*_PUBLIC_HOST` variables carry the bare hostname (no scheme or port) and become the HTTPRoute hostnames on the plane gateways.
 :::
 
 ---
@@ -219,6 +219,7 @@ helm install amp \
   --set agentManagerService.config.oidc.clientSecret="${AMP_API_CLIENT_SECRET}" \
   --set agentManagerService.config.thunder.clientSecret="${AMP_SYSTEM_CLIENT_SECRET}" \
   --set agentManagerService.config.openChoreo.baseURL="http://openchoreo-api.openchoreo-control-plane.svc.cluster.local:8080" \
+  --set agentManagerService.config.tlsEnabled=true \
   --set agentManagerService.replicaCount=2 \
   --set agentManagerService.autoscaling.minReplicas=2 \
   --set console.replicaCount=2 \
@@ -249,6 +250,10 @@ kubectl wait --for=condition=Available \
 ```
 
 `agentManagerService.config.thunder.clientSecret` can instead be supplied through `agentManagerService.config.thunder.existingSecret` (key `thunder-client-secret`) if you prefer to keep it out of Helm release history.
+
+:::warning `tlsEnabled` decides which agent invoke URL the console publishes
+`agentManagerService.config.tlsEnabled` defaults to `false`, and it is what selects between an Environment's `http` and `https` endpoint variants when the API builds an agent's invoke URL. Left at the default on an HTTPS platform, the console publishes an `http://` URL — which the browser then refuses to call from an HTTPS page, showing the request as `blocked:mixed-content` with no error anywhere in the platform. The agent itself is healthy and reachable; only the published URL is wrong.
+:::
 
 :::warning Two values that look redundant but are not
 `agentManagerService.autoscaling.minReplicas=2` is what actually gives the API two replicas. Autoscaling is **on by default** for this component with `minReplicas: 1`, so the HPA takes ownership of the replica count and scales `replicaCount=2` straight back down to one — silently, within seconds. The console has autoscaling off by default, which is why `replicaCount` alone appears to work there.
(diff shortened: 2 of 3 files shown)


RAVEENSR:
Addressed the review comment, and replaced one workaround with the actual fix.

**`AGENTS_DOMAIN` (review comment):** the prerequisite now states what the variable is and how to derive it — "the domain deployed agents are served on, `agents.<your base domain>`" — rather than pointing at a section that does not define it. The production page exports it in Plan Your Deployment; the k3d page never renders these steps, since the gateway `vhost`/`hostname` settings are inside the production-only variant of the partial.

**OpenSearch password — now fixed at the source instead of worked around.** The first revision had the guide `sed`-substitute the password while applying `deployments/values/oc-collector-configmap.yaml`. That treated the symptom. The manifest is a fork of the `observability-tracing-opensearch` module's own collector config, and the fork replaced two env-var references with literals:

```yaml
# upstream observability-tracing-opensearch 0.4.1
basicauth/opensearch:
  client_auth:
    username: ${env:OPENSEARCH_USERNAME}
    password: ${env:OPENSEARCH_PASSWORD}
```

The module already injects both from the `opensearch-admin-credentials` Secret via its default `opentelemetry-collector.extraEnvs`, and no install path overrides that — verified by rendering the chart, the env vars are present in the collector pod on every path. So this commit restores the two references, and the guide reverts to a plain `kubectl apply` with no substitution and no admonition.

This is also safer on the local/k3d paths, which apply the same manifest: there the placeholder happened to match the seeded password, which is exactly why the mismatch was invisible until an install changed it. The Secret holds that same value locally, so the env-var form works there too.

Remaining items from the same validation run are being raised separately, since they are chart and service fixes rather than documentation: the agent traits hardcoding a gateway namespace that contradicts `apiGateway.namespace`'s own default, per-replica gateway connection status, `gateways.vhost` not being reconcilable, the registration token Secret surviving `helm uninstall`, and the evaluation-job NetworkPolicy's gateway egress selector.

RAVEENSR:
Reworked the `LimitRange` guidance — the previous revision presented resource values I had not measured as though they were recommendations.

**What stands:** the `LimitRange` itself is required, not a workaround. A ResourceQuota tracking `requests.*`/`limits.*` rejects any pod that omits them, and build pods declare none, so the quota alone fails every build at admission. Verified twice on a cluster: with a bare probe pod, and with a real build that went terminally `Failed` in 15s at its first step.

**Why a LimitRange rather than resources on the build templates.** The rejection names three containers — `must specify limits.cpu for: init,main,wait`. Only `main` comes from the workflow templates in this repo; `init` and `wait` are injected by Argo, and their resources come from the workflow controller's configuration, which the workflow-plane chart owns rather than this one. So declaring resources on the build steps would not make the quota usable on its own, while one `LimitRange` covers all three from an object the operator controls. Adding requests to the step containers is still worth doing separately — it moves builds off BestEffort QoS and gives the quota meaningful accounting — but it is an improvement, not what unblocks this.

**What changed in the doc:**

- the values are labelled placeholders rather than recommendations, with an inline `# Replace with your measured values`;
- added the procedure to derive them — run a build with no `LimitRange` and watch `kubectl top pod -n workflows-default --containers`, then set `defaultRequest` near steady state and `default` above peak;
- added a warning that limits set too low get the step OOMKilled or CPU-throttled into a timeout, which reads as a broken build rather than a resource cap, with the guidance to prefer generous limits and a smaller quota since the quota is what actually bounds total consumption;
- explained the `init`/`main`/`wait` detail so the error message is recognisable.

If anyone has real figures for what a buildpack build consumes, I would rather put those in than ask every reader to measure it — happy to update.

menakaj (approved):

PR #1501, opened 2026-08-05
Fix install-guide traps found in a from-scratch EKS run

jhivandb:
## Purpose

A from-scratch install on EKS (external PostgreSQL on RDS, Let's Encrypt DNS-01 via Route 53) following the `next` on-your-environment guide verbatim hit several failures the guide doesn't cover. Each change below corresponds to a failure actually reproduced during that run — three break the guide's own commands, the rest are environment traps the production path steers readers into.

## Changes

**Guide commands that fail as written**

- **`VERSION="0.0.0-dev"` admonition** — the placeholder is swapped for a real tag at release, but readers of the `next` docs hit it live, and it exists as neither a GHCR chart tag nor an `amp/v*` git tag, so the very first `helm pull` fails. Added a caution with a one-liner to find the latest nightly tag.
- **OpenBao readiness wait** — `kubectl wait --for=condition=Ready=false pod/openbao-0` is satisfied before the container starts (and a sealed OpenBao never goes Ready), so the `bao operator init` exec races it and dies with `container not found ("openbao")`. Replaced with a poll on `bao status` exit codes (0 = unsealed, 2 = sealed but responding).
- **Helm 4** — satisfies the documented "v3.12+" but its server-side apply conflicts with charts whose controllers rewrite their own webhook `caBundle` (OpenBao's injector, cert-manager's cainjector) on any re-run: `Apply failed with 1 conflict ... "vault-k8s" ... clientConfig.caBundle`. Since the guide's own troubleshooting recommends re-running installs, pinned the prerequisite to v3 and documented the `--server-side=false` escape hatch.

**Environment traps on the production path**

- **Managed PostgreSQL** — RDS/Cloud SQL/Azure admin users are not superusers, and PostgreSQL 16+ rejects `CREATE DATABASE ... OWNER thunder` with `must be able to SET ROLE "thunder"` until the role is granted to the admin. Added the `GRANT ... TO CURRENT_USER` step.
- **Registry requirements** — build workflows push `<workflow-run-name>-image` (a new repository per run, so push-to-create is mandatory) using a static `.dockerconfigjson` (so expiring credentials can't work). ECR fails both, and the failure only surfaces on the first build. Documented the two requirements, which registries satisfy them, and the no-auth caveat of a quick in-cluster Distribution deploy.
- **Default StorageClass prerequisite** — stock EKS satisfies "Required" with a `gp2` class bound to the removed in-tree provisioner; PVCs accept and never bind. Qualified the prerequisite.

**Fail-fast additions**

- DNS delegation pre-check (via a public resolver, not the zone's own NS) before the DNS-01 issuer — an undelegated zone otherwise leaves every certificate hanging unissued with no error.
- LoadBalancer smoke test in the prerequisites, so a missing LB controller surfaces before the install rather than at Step 6.
- Note that cert-manager picks up ambient credentials (IRSA/Pod Identity) on EKS as the safer alternative to a static access key.

## Verification

- `npm run build` in `documentation/` passes on the branch.
- Every runtime failure above was reproduced and its fix validated during the EKS install run (2026-08-05, EKS 1.34 / Helm 4.2.0 / RDS PostgreSQL 17.10 / OpenBao chart 0.25.6).

## Summary by CodeRabbit

* **Documentation**
  * Added registry requirements for build workflows, including repository creation on push and static, non-expiring credentials.
  * Clarified supported registry options and authentication requirements for Distribution deployments.
  * Expanded environment setup guidance for storage, Helm 3, load balancers, DNS delegation, credentials, and PostgreSQL permissions.
  * Updated OpenBao readiness guidance to use status checks.
  * Added guidance for validating LoadBalancer provisioning and DNS configuration.

### DIFF
--- a/documentation/docs/getting-started/on-your-environment.mdx
+++ b/documentation/docs/getting-started/on-your-environment.mdx
@@ -45,7 +45,7 @@ The main flow assumes you control a DNS zone. If you are evaluating on a cluster
 | CPU per node | 4 cores |
 | RAM per node | 8 GB |
 | LoadBalancer support | Required |
-| Default StorageClass | Required |
+| Default StorageClass | Required, and it must actually provision volumes. Stock EKS ships a default `gp2` class bound to the **removed** in-tree provisioner: PVCs are accepted and stay `Pending` forever. Install the EBS CSI driver and make a CSI-backed class (e.g. `gp3`) the default |
 | Node kernel (build nodes) | Linux **6.3+** for user-namespaced agent builds (idmapped-mount support). On older kernels (Ubuntu 22.04's 5.15, Amazon Linux 2's 5.10) set `buildWorkflows.userNamespaces=false` on the Platform Resources chart — see [Troubleshooting](#build-pods-fail-with-mount_attr_idmap--idmap-mounts-errors) |
 | NetworkPolicy-enforcing CNI | Recommended. The Observability and Evaluation extension charts each ship a `NetworkPolicy` (collector ingress, evaluation-job egress) — inert on a CNI that doesn't implement the NetworkPolicy API. EKS's default VPC CNI, AKS, and legacy GKE don't enforce it out of the box; enable your provider's policy add-on (or install Calico/Cilium) if you want these to actually take effect. k3s (local/CI) enforces it by default. |
 
@@ -61,11 +61,23 @@ The main flow assumes you control a DNS zone. If you are evaluating on a cluster
 | Tool | Version | Purpose |
 |------|---------|---------|
 | [kubectl](https://kubernetes.io/docs/tasks/tools/) | v1.32+ | Kubernetes CLI |
-| [Helm](https://helm.sh/docs/intro/install/) | v3.12+ | Kubernetes package manager |
+| [Helm](https://helm.sh/docs/intro/install/) | v3.12+, **v3 only** | Kubernetes package manager |
 | curl / dig | — | DNS resolution of LoadBalancer hostnames |
 
 ```bash
 kubectl version --client && helm version
+```
+
+:::caution Helm 4 is not supported by this guide
+Helm 4 applies charts with Kubernetes server-side apply, where every field has an owner. Several charts below (OpenBao, cert-manager) have controllers that rewrite their own webhook `caBundle` at runtime under a different field manager, so any **re-run** of an install command — including the retries this guide's own troubleshooting recommends — fails with `Apply failed with 1 conflict`. Use Helm 3, or append `--server-side=false` to every `helm install`/`helm upgrade` in this guide.
+:::
+
+Verify LoadBalancer provisioning works before starting — the alternative is discovering it an hour into the install, when the first plane gateway hangs `<pending>`:
+
+```bash
+kubectl create deploy lbtest --image=nginx && kubectl expose deploy lbtest --port=80 --type=LoadBalancer
+kubectl wait --for=jsonpath='{.status.loadBalancer.ingress}' svc/lbtest --timeout=5m && echo "LoadBalancer OK"
+kubectl delete svc/lbtest deploy/lbtest
 ```
 
 ### Permissions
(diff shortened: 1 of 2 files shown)


documentation/docs/getting-started/on-your-environment.mdx
```diff
@@ -45,7 +45,7 @@ The main flow assumes you control a DNS zone. If you are evaluating on a cluster
 | CPU per node | 4 cores |
 | RAM per node | 8 GB |
 | LoadBalancer support | Required |
-| Default StorageClass | Required |
+| Default StorageClass | Required, and it must actually provision volumes. Stock EKS ships a default `gp2` class bound to the **removed** in-tree provisioner: PVCs are accepted and stay `Pending` forever. Install the EBS CSI driver and make a CSI-backed class (e.g. `gp3`) the default |
```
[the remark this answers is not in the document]
jhivandb: Fixed in 45148a5d — the prerequisite now distinguishes the two EKS cases: 1.30+ clusters ship no default StorageClass at all, and the stock gp2 class (still default on upgraded clusters) requires the EBS CSI driver because the in-tree provisioner it references was removed in Kubernetes 1.27. Both paths end at the same instruction: install the EBS CSI driver and mark a CSI-backed class as default.

documentation/docs/getting-started/on-your-environment.mdx
```diff
@@ -133,6 +145,14 @@ export INSTRUMENTATION_URL="http://default-default.gateway.localhost:19080/otel"
 The certificates below are single-level wildcards (`*.<base>`), so every management hostname must sit directly under the base domain — no second-level names like `api.amp.<base>`.
 :::
 
+:::caution Reading the unreleased (next) version of this guide?
+`0.0.0-dev` is a placeholder that is swapped for the real tag at release — it is **not itself a published tag**, so every `helm pull` and values-file download below will fail with it. Use the latest nightly instead, tagged `0.0.0-dev-<YYYYMMDD>`:
+
+```bash
+git ls-remote --tags https://github.com/wso2/agent-manager 'refs/tags/amp/v0.0.0-dev-*' | tail -1
+```
```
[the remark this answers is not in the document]
jhivandb: Fixed in 45148a5d — the snippet now exports the parsed VERSION (tag name extracted, leading v stripped, latest by date) and fails visibly via ${VERSION:?} when no nightly tag matches, rather than exit 1, which would close an interactive shell. Verified against the live repo: it currently resolves to 0.0.0-dev-20260805.

documentation/docs/getting-started/on-your-environment.mdx
```diff
@@ -133,6 +145,16 @@ export INSTRUMENTATION_URL="http://default-default.gateway.localhost:19080/otel"
 The certificates below are single-level wildcards (`*.<base>`), so every management hostname must sit directly under the base domain — no second-level names like `api.amp.<base>`.
 :::
 
+:::caution Reading the unreleased (next) version of this guide?
```
RAVEENSR: I don't think this section is necessary. Because this guide will always have a released version associated with it when a doc release is triggered and we are not encouraging people to read the next doc.

jhivandb: Problem is the docsite defaults to next when you visit it, most people r just gonna hop onto the site and run the commands they see. Tho I can ack this might not be the best way to show it

RAVEENSR: no, doc site always defaults to the latest released version. check this https://wso2.github.io/agent-manager/ when you click the **Get started** it goes to the latest released version.

RAVEENSR: However just now I discovered that when you visit this link https://wso2.github.io/agent-manager/ and click on the top left documentation icon, it goes to the next. This is an issue in doc site. We need to fix it. So let's remove the section and fix the redirection correctly.

jhivandb: Sounds good, will do

jhivandb: Removed the comment.

Regarding the bug, its because at some point we've selected "next" and the docs app has saved this to localstorage. So when we click the docs icon it redirects us to next. It shouldn't be a problem for users

RAVEENSR (approved):

PR #1716, opened 2026-08-22
Route the evaluation job through the gateway's internal URL

jhivandb:
## Purpose
The monitor evaluation job received the gateway's **public vhost** as `LLM_API_BASE`, so an in-cluster workflow pod had to hairpin out through the public ingress and back in to reach an LLM proxy sitting in the same cluster.

The infrastructure for the direct path already existed but nothing used it:

- `models.Gateway.RuntimeURL` holds the in-cluster address (e.g. `http://api-platform-acme-dev-gw-gateway-gateway-runtime.acme-dev:22893`), computed by the gateway extension chart and registered at bootstrap. It was validated on register/update and returned by the API, but **no code built a URL from it**.
- The evaluation job's NetworkPolicy already allows egress to the gateway namespace on port **22893** — the runtime Service port (`deployments/helm-charts/wso2-amp-evaluation-extension/templates/evaluation-job-networkpolicy.yaml`). The public-vhost path only worked via the local-dev `devEgress` ipBlock on the node network (19080).

## Goals
Give the evaluation job the gateway's in-cluster runtime URL instead of the public vhost, so LLM-judge traffic stays inside the cluster and matches the egress the job is actually granted.

Agent env-var injection, config responses, and MCP proxies are unchanged — they keep the public URL, which must stay identical to the identity resource-server identifier.

## Approach
`buildProxyURL` was the single URL builder for LLM proxies and had no counterpart, which made it easy to reach for the wrong address. It is now an explicit pair:

- **`buildProxyURL` → `buildPublicProxyURL`** — unchanged behavior, 6 call sites (agent env injection, config responses, proxy provisioning).
- **`buildInternalProxyURL`** (new) — builds from `gateway.RuntimeURL`. No scheme defaulting: `validateGatewayRuntimeURL` already requires an explicit `http`/`https`, and forcing `https` would break a plaintext in-cluster address.

`monitorExecutor.resolveProxyURL` → `resolveInternalProxyURL`, now returning `buildInternalProxyURL(...)`.

**Fails closed.** A gateway with no registered `runtimeUrl` returns `errGatewayRuntimeURLUnregistered` (a sentinel, matching the existing gateway-resolution errors in `services/gateway_roles.go`) rather than silently falling back to the vhost — a fallback would route the job out through an ingress its own NetworkPolicy denies, turning a provisioning fault into a confusing network error.

**Why no Host header is needed:** the gateway only applies a host constraint when the deployed artifact carries a `vhost`. For LLM proxies that value comes from `LLMProxyConfig.Vhost`, which has no writer anywhere in AMS — so every LLM proxy deploys with an empty vhost and the gateway routes it on the context path alone. (MCP deliberately differs: `mcpProxyServingBase` honors a per-proxy vhost override, which is why no MCP counterpart is added here.)

## User stories
As a platform operator, when a monitor runs LLM-judge evaluators, the evaluation job reaches the LLM proxy over cluster-internal networking instead of egressing to the public ingress and back.

## Release note
Monitor evaluation jobs now reach the LLM gateway over its in-cluster runtime address instead of the public vhost. A monitor whose gateway has no registered `runtimeUrl` now fails its run instead of attempting the public URL.

## Documentation
N/A — no user-facing API, CLI, or console surface changes. The behavior is internal routing between the evaluation job and the gateway.

## Training
N/A — no training content covers evaluation-job network routing.

## Certification
N/A — internal routing detail, not covered by certification exams.

## Marketing
N/A — internal reliability fix with no promotable surface.

## Automation tests
 - Unit tests
   > `services` package, all passing via `make test-unit` (full tier green).
   > - `TestBuildInternalProxyURLUsesRuntimeURL` — runtime URL with and without a context path; asserts the `http` scheme survives verbatim and is not upgraded to `https`.
   > - `TestBuildInternalProxyURLRejectsMissingRuntimeURL` — empty and whitespace-only `RuntimeURL` both return `errGatewayRuntimeURLUnregistered` (asserted with `errors.Is`, not string matching).
   > - `TestAnchoring_MonitorRunAgainstDeployedProxy` — existing gateway-anchoring test updated to expect the runtime URL; new subtest `fails when the anchored gateway has no runtimeUrl` covers the fail-closed path end-to-end through `resolveInternalProxyURL`.
   > - `TestBuildPublicProxyURLUsesVhost` and the MCP builder tests are unchanged and still green, pinning the no-regression guarantee for agent env injection.
 - Integration tests
   > `tests/monitor_executor_test.go::TestExecuteMonitorRun_LLMCredentials` updated: the seeded gateway now carries a `runtime_url`, and the assertion on the workflow's `llmApiBase` parameter expects the runtime URL + proxy context path. Compiles clean under `-tags=integration`; **not executed** — requires a provisioned test database, so it needs a CI run.

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? **yes**
 - Ran FindSecurityBugs plugin and verified report? **no** — FindSecurityBugs is a Java/SpotBugs plugin and does not apply to this Go module. `make lint` (golangci-lint, the repo's Go gate) reports 0 issues.
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? **yes** — the change moves a URL only. The LLM API key path is untouched: it still arrives via the existing `SecretReference` remoteRef, never inlined.

## Samples
N/A — no sample changes; the behavior is internal to monitor execution.

## Related PRs
None. Builds on the existing `runtimeUrl` registration work (commits `3466bdcbc`, `31b1cd4b5`, migration `038_gateway_runtime_url.go`), which added and populated the field but had no consumer.

## Migrations (if applicable)
No new migration. **This PR does change the blast radius of an existing one, and reviewers should weigh it:**

Migration 038 backfills `runtime_url` only for gateways whose name starts with `api-platform-`. A custom-named gateway registered before that migration keeps a NULL `runtime_url`, and its monitors will now **fail** rather than fall back to the public vhost. The repair path already exists — the gateway extension chart's bootstrap job PUTs `runtimeUrl` on every upgrade (`sync_runtime_url` in `gateway-bootstrap-job.yaml`) — so a chart upgrade fixes affected gateways. Operators running a custom-named gateway without a chart upgrade should confirm `GET /gateways` reports a non-empty `runtimeUrl` before deploying this.

Failing closed here is deliberate: the fallback would produce an opaque network error from inside the job rather than a clear provisioning error at resolution time.

## Test environment
- Go 1.25.11 (darwin/arm64), macOS 15 (Darwin 25.0.0)
- `make lint` (golangci-lint v2.1.6) — 0 issues
- `make test-unit` — full unit tier passing
- `go vet -tags=integration ./tests/...` — clean
- No live cluster verification: the local k3d cluster was unreachable (TLS verify failure) during this work. The claim that the gateway routes LLM proxies on the runtime address without a Host header is derived from configuration (empty artifact vhost + the NetworkPolicy's existing 22893 allowance), **not** from an observed request. Worth confirming on a live environment before merge.

## Learning
The non-obvious part was establishing that no `Host` header workaround is needed. The public path routes through kgateway, whose HTTPRoute matches on `hostnames: [<vhost>]` — which suggested that bypassing it to hit the runtime Service directly would 404. Tracing the deploy payload showed the artifact-level vhost is the only host constraint the gateway itself applies, and `LLMProxyConfig.Vhost` has no writer in AMS, so LLM proxies are always deployed host-unconstrained. The evaluation job's NetworkPolicy already allowing port 22893 confirmed the direct path was the originally intended design.

## Summary by CodeRabbit

- **New Features**
  - Improved proxy URL handling by distinguishing public gateway addresses from internal runtime addresses.
  - Internal services now use the configured gateway runtime URL, including custom schemes, ports, and context paths.
  - Missing runtime URLs fail safely instead of falling back to public addresses.

- **Bug Fixes**
  - Monitor and LLM proxy integrations now resolve to the appropriate internal or public gateway URL, improving connectivity in deployed environments.

### DIFF
--- a/agent-manager-service/services/agent_configuration_proxy_url_test.go
+++ b/agent-manager-service/services/agent_configuration_proxy_url_test.go
@@ -19,20 +19,22 @@ package services
 import (
 	"testing"
 
+	"github.com/google/uuid"
+
 	"github.com/stretchr/testify/require"
 
 	"github.com/wso2/agent-manager/agent-manager-service/models"
 )
 
-func TestBuildProxyURLUsesVhost(t *testing.T) {
+func TestBuildPublicProxyURLUsesVhost(t *testing.T) {
 	// RuntimeURL present but must be ignored: every agent gets the public URL.
 	gateway := &models.Gateway{
 		Vhost:      "https://dev-acme.gateway.example.com",
 		RuntimeURL: "http://api-platform-acme-dev-gw-gateway-gateway-runtime.acme-dev:22893",
 	}
 	contextPath := "/llm/proxy"
-	require.Equal(t, "https://dev-acme.gateway.example.com/llm/proxy", buildProxyURL(gateway, &contextPath))
-	require.Equal(t, "https://dev-acme.gateway.example.com", buildProxyURL(gateway, nil))
+	require.Equal(t, "https://dev-acme.gateway.example.com/llm/proxy", buildPublicProxyURL(gateway, &contextPath))
+	require.Equal(t, "https://dev-acme.gateway.example.com", buildPublicProxyURL(gateway, nil))
 }
 
 func TestBuildMCPProxyURLUsesGatewayVhost(t *testing.T) {
@@ -61,12 +63,12 @@ func TestBuildMCPProxyURLPrefersProxyVhostOverride(t *testing.T) {
 		buildMCPProxyURL(gateway, models.MCPProxyConfig{Vhost: &empty}))
 }
 
-func TestBuildProxyURLDefaultsBareVhostToHTTPS(t *testing.T) {
+func TestBuildPublicProxyURLDefaultsBareVhostToHTTPS(t *testing.T) {
 	// Gateways registered with a scheme-less vhost must still yield absolute URLs.
 	gateway := &models.Gateway{Vhost: "gw.example.com"}
 	contextPath := "/llm/proxy"
-	require.Equal(t, "https://gw.example.com/llm/proxy", buildProxyURL(gateway, &contextPath))
-	require.Equal(t, "https://gw.example.com", buildProxyURL(gateway, nil))
+	require.Equal(t, "https://gw.example.com/llm/proxy", buildPublicProxyURL(gateway, &contextPath))
+	require.Equal(t, "https://gw.example.com", buildPublicProxyURL(gateway, nil))
 }
 
 func TestBuildMCPProxyURLDefaultsBareGatewayVhostToHTTPS(t *testing.T) {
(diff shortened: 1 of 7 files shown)


RAVEENSR:
I hit this same bug while installing 1.0.0-RC1 on a VM and had started an equivalent fix (#1717) before spotting this one — closing mine in favour of it, since the public/internal rename across all call sites and the migration-038 analysis are both better than what I had. Passing on the verification I'd already done, since the integration test here hasn't been executed yet.

**Verified on a GCP VM running 1.0.0-RC1 (simple install), with the evaluation-job egress policy exactly as shipped** — no ipBlock workaround in place.

Built this branch, deployed it as `amp-api`, and the next scheduled monitor run was handed the in-cluster address:

```
LLM_API_BASE = http://api-platform-default-default-gw-gateway-gateway-runtime.default-default:22893/<proxy-uuid>
```

That particular run selected `0 trace(s)` (the agent had been idle, so its 5-minute window was empty), so to exercise the judge path itself I ran the `amp-evaluation-monitor:v1.0.0-RC1` image directly: carrying the job's two selector labels so the NetworkPolicy actually applies, pointed at the URL this branch minted, authenticating with the real gateway key from an existing per-run `*-llm-proxy-credentials` secret, and calling through `any_llm` the way `base.py` does. The model replied normally — pod → policy → in-cluster gateway → provider → response.

An earlier run against the same URL shape (my build, same address) completed `status=SUCCESS` with Accuracy 3/3 (mean 1.000), Reasoning Quality 3/3 (0.833) and Groundedness 3/3 (0.333), where the same monitor had previously skipped every evaluator on every trace with `Connection error`.

Supporting measurements, in case they're useful for the description:

- From a pod carrying both selector labels, the in-cluster gateway on `:22893` is reachable 6/6 while the public host is refused 6/6 (`ECONNREFUSED`). An unlabeled control pod reaches both.
- The runtime route matches on path alone, no public `Host` header needed: the real proxy path returns `401` with a deliberately wrong key, a bogus path `404`.
- The original failure reproduces from inside the monitor image as `openai.APIConnectionError` wrapping `[Errno 111] Connection refused`.

Two traps worth knowing for anyone re-testing this: NetworkPolicy egress is matched **post-DNAT**, which is why the same URL is permitted where it lands on an in-cluster ingress and refused where it lands on a host-network container; and enforcement is programmed a second or two **after** pod start, so a single quick `curl` in a fresh pod succeeds and looks like proof the path is healthy.

**One question on the fail-closed choice.** An AI gateway registered as `EGRESS` that runs outside the cluster — the deployment the production guide describes for a private model endpoint, connecting back through the control-plane gateway — has no in-cluster address by definition, so its monitors will now fail permanently rather than degrade. The argument for fail-closed still holds for the common case: the vhost would be refused by the policy anyway, and a named error beats a mystery `Connection error`. But if an operator has deliberately opened egress to that external gateway, this removes a setup that was working. Worth deciding explicitly rather than by omission — either a vhost fallback for gateways with no `runtimeUrl`, or a line in the release note saying externally-hosted egress gateways aren't supported for monitors.

I did not get to observe the fail-closed symptom on a live cluster (it needs a `runtime_url` cleared in the DB, which I didn't want to do on a shared box); the unit coverage here looks right for the code path itself.

RAVEENSR (approved):

PR #1756, opened 2026-08-26
Upgrade api gateway helm chart to v1.2.2 and fix guardrail configuration

menakaj:
## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary by CodeRabbit

- **New Features**
  - Added support for raw TOML gateway configuration and additional environment variables for gateway deployments.
  - Improved handling of empty values in nested policy configuration arrays.

- **Bug Fixes**
  - Deprecated gateway environment configuration is now rejected to prevent invalid deployments.

- **Documentation**
  - Updated installation guidance and examples for the latest gateway chart version.
  - Clarified configuration options and Secret-backed environment injection.

- **Chores**
  - Updated quick-start and setup scripts to use gateway chart version 1.2.2.

### DIFF
--- a/console/workspaces/libs/shared-component/src/utils/policyParameterEditor/schemaUtils.ts
+++ b/console/workspaces/libs/shared-component/src/utils/policyParameterEditor/schemaUtils.ts
@@ -492,6 +492,20 @@ export function coerceValuesToSchemaTypes(
     stripExclusiveBranchKeys(branches, result);
   }
 
+  // Never send unset optional fields to the gateway: an explicit empty
+  // string/array/object for an optional param (e.g. an unconfigured
+  // endpoint or API key) fails config.toml parsing downstream, whereas
+  // omitting the key lets the policy fall back to its own default.
+  // Required keys are left as-is so blank-but-required fields still
+  // surface as a validation error instead of silently vanishing.
+  const requiredKeys = new Set(schema.required ?? []);
+  Object.keys(result).forEach((key) => {
+    if (requiredKeys.has(key)) return;
+    if (!hasMeaningfulValue(result[key])) {
+      delete result[key];
+    }
+  });
+
   return result;
 }
 
--- a/console/workspaces/libs/shared-component/src/utils/policyParameterEditor/schemaUtils.ts
+++ b/console/workspaces/libs/shared-component/src/utils/policyParameterEditor/schemaUtils.ts
@@ -442,6 +442,7 @@ function stripExclusiveBranchKeys(
 export function coerceValuesToSchemaTypes(
   schema: ParameterSchema,
   values: ParameterValues,
+  options?: { isArrayItem?: boolean },
 ): ParameterValues {
   if (schema.type !== "object" || !schema.properties) return values;
 
@@ -474,9 +475,14 @@ export function coerceValuesToSchemaTypes(
           typeof item === "object" &&
           item !== null
         ) {
+          // Array items are coerced but not stripped of empty keys: a
+          // half-filled row (e.g. {name: "", value: ""}) must keep every
+          // item-schema key present, or the backend receives a bare `{}`
+          // with none of the row's fields.
           return coerceValuesToSchemaTypes(
             propSchema.items!,
             item as ParameterValues,
+            { isArrayItem: true },
           );
         }
         return coerceValue(item, propSchema.items!);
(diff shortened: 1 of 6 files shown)


jhivandb:
## Review notes

Reviewed `ba130ac0e`. The Helm mechanism was verified end-to-end rather than reasoned about: pulled gateway charts `1.2.0` and `1.2.2` from ghcr.io, rendered the extension chart with populated `config_toml` + `systemExtraEnv`, and fed the generated inner `values.yaml` into the real downstream chart. **The core approach is correct** — `nindent 8` block-scalar nesting round-trips exactly, `{{ env "..." }}` tokens pass through un-templated as intended, and only the two deployments receiving `extraEnv` mount the config ConfigMap, so injecting into both is right. The 1.2.0→1.2.2 values diff is purely additive, and 1.2.2's image defaults really are `1.2.0`, so the comments in `install.sh` / `env.sh` / docs hold.

### Findings

**1. `console/.../policyParameterEditor/schemaUtils.ts:504` — medium**

Pruning every "non-meaningful" optional value makes *explicitly cleared* indistinguishable from *never set*:

- An optional scalar with a non-empty schema `default` can't be cleared — the blank is dropped from the PUT, the gateway re-applies the default, and reopening the drawer re-seeds it via `initializeDefaultValues`. The clear silently round-trips back to the default.
- `hasMeaningfulValue([])` is `false`, so emptying a list-shaped param (an allow-list like `allowedTools` / `allowedTopics` on a filtering guardrail) drops the key and the gateway applies the policy default instead of the empty list — potentially fail-open where the user asked for fail-closed.

Suggestion: key on "user touched and cleared" rather than emptiness, or exempt `type: array`.

**2. Same file — low**

Inside array items the strip runs recursively, so a half-filled list now sends `{}` entries. Concrete: user clicks "Add item", fills row 1, leaves row 2 blank. `hasMeaningfulValue` uses `.some()`, so the array survives, but row 2 — previously `{name:"",value:""}` — is reduced to `{}` and sent with none of the item schema's keys present.

**3. `deployments/helm-charts/wso2-amp-api-platform-gateway-extension/values.yaml:187` — low**

The new comment says "Whole `[table]` sections work here too", but drops the warning the downstream chart's own `config_toml` docs carry: *do not redefine a table the chart already emits (TOML duplicate-table error)*. This chart already emits `[policy_configurations.*]`, `[controller.*]`, `[router]`. Someone following the comment and pasting e.g. `[policy_configurations.ratelimit_v1]` gets a hard duplicate-table error at config load — gateway-runtime crash-loops, with nothing in the values file to warn them.

**4. `values.yaml:254` — low**

`systemEnv` is removed with no deprecation path. There's no `values.schema.json`, so an operator upgrading with `apiGateway.config.systemEnv:` still in their overrides gets a clean `helm upgrade` and zero signal that it needs migrating to `config_toml` + `systemExtraEnv`. (Nothing inside this repo still references it, and per the downstream README a plain `APIP_GW_*` var with no matching `config.toml` token was already ignored — so this is a migration-signal gap, not a functional regression.) A one-liner would turn a silent no-op into an actionable error:

```
{{- if .Values.apiGateway.config.systemEnv }}{{ fail "apiGateway.config.systemEnv has been replaced by config_toml + systemExtraEnv" }}{{- end }}
```

**5. `values.yaml:106` — low**

`chartVersion: "1.1.0"`, commented "must match what gateway-operator is configured for", is the one spot this PR's `1.2.0 → 1.2.2` sweep missed. It's referenced by no template in the chart (the operator's `gateway.helm.chartVersion` is set in `ensure-gateway-operator.sh` / `install.sh`), so it misinforms rather than breaks. Worth bumping or deleting.

**6. `values.yaml:232` — low**

The `__` escaping in the example env names (`APIP_GW_GRANITE_GUARDIAN__API__KEY`, `APIP_GW_AWSBEDROCK_SECRET__ACCESS__KEY`) is a leftover from the deleted koanf key-derivation rule and no longer means anything — `{{ env "NAME" }}` is a literal lookup. It's also inconsistent with the `systemExtraEnv` example just below, which uses plain `APIP_GW_AZURECONTENTSAFETY_KEY`. Copying the token verbatim but writing the "natural" single-underscore name in `systemExtraEnv` yields an unresolved token with no name-mismatch diagnostic.

**7. `console/licenses/THIRD_PARTY_LICENSES.md` — low, scope**

Looks unrelated to a gateway chart bump — no `package.json` or lockfile changed in this diff. It also self-describes as unverified ("`@wso2/oxygen-ui*` license was not verified directly", "No `node_modules` was present locally at audit time… Re-verify with `license-checker`"), which is an awkward property for a compliance doc landing in the repo; the sibling `agent-manager-service/licenses/THIRD_PARTY_LICENSES.md` carries no such caveats. Suggest splitting it into its own PR after verifying against a real install.

### Verified clean

- Rendered the chart under both default and populated values: `config_toml` and both `extraEnv` blocks indent correctly and the inner `values.yaml` re-parses.
- `gateway.config_toml`, `gateway.controller.deployment.extraEnv`, and `gateway.gatewayRuntime.deployment.extraEnv` all exist in 1.2.2; the 1.2.0→1.2.2 diff only adds keys (`config_dump`, `trafficLogVolume`, `routerAdminEnabled`).
- Version pins updated consistently across `install.sh`, `env.sh`, and current docs; `versioned_docs` snapshots correctly untouched.

One note for the PR description: `config_toml` and both `extraEnv` keys already existed in chart `1.2.0`, so the chart bump isn't strictly required by this feature.

jhivandb (approved):

PR #1778, opened 2026-08-31
Add full-URL registration for env-Thunder, on-prem and SaaS

### LINKED ISSUE #1770: [Improvement]: For Env-thunder support cloud-provisioned URLs, not just on-prem handles

Area/Other (Anything else which does not match above categories)

Aspect/Identity & Access (Identities, identity providers, roles, permissions, tenancy)

### Suggested Improvement

**Problem**

The database only saves a short handle for each environment's Thunder identity server (e.g. `prodction-idp`). Every reader rebuilds the full address by gluing that handle onto one shared domain setting. That works on-prem, where every environment lives under one domain — it breaks for cloud/SaaS, where environments can live under different domains with no single pattern to reconstruct from.

**Fix**

Add a `thunder_url` column that stores the full resolved address directly, as the single source of truth. Two ways to register it on the same endpoint:
- **On-prem** — give a handle (or let it auto-generate), server computes and stores the URL, exactly as before.
- **SaaS** — control plane hands over the real URL directly, no handle involved.

Every reader now uses the stored URL instead of recomputing it.

**Bug caught along the way**

Handle-less (SaaS) rows were colliding with each other on save, because an omitted handle was being stored as an empty string instead of a real database `NULL`. Fixed by making the field nullable properly; verified against a real database.

### Related Issues

_No response_

yashed:
## Purpose
- Resolved : https://github.com/wso2/agent-manager/issues/1770
Env-Thunder (each environment's dedicated Thunder ID instance) URLs were only ever stored as a short handle, reconstructed into a full address at read time using one global domain config. That breaks for a cloud/SaaS control plane that provisions environments under different domains, since there's no single pattern to reconstruct from.

## Goals
- Store the full, resolved env-Thunder origin instead of reconstructing it from a handle on every read.
- Support two registration paths on the same endpoint: on-prem (handle, server computes the URL) and SaaS/control-plane (caller supplies the full URL directly).
- Keep on-prem behavior fully backward compatible, no change to the existing handle flow.
- Close the SSRF exposure a caller-supplied URL introduces, since that value now gets dialed later, not just stored.

## Approach
- New `thunder_url` column on `env_thunder_urls`, the single source of truth every reader now uses.
- `thunder_handle` becomes nullable (`*string` in Go, not `string`), so an omitted handle writes a real SQL `NULL`. Needed so multiple handle-less SaaS rows don't collide on the handle uniqueness constraint.
- `PUT .../thunder-url` accepts `handle` or `url`, never both. Sending both is rejected with 400.
- On-prem path unchanged in behavior: validate/generate the handle, then compute and store the origin server-side via `ThunderOriginFromHandle`.
- SaaS path: validate the URL's shape (no userinfo/path/query, http/https only) and run it through the existing `ssrf.ValidateURL`, then store it verbatim with no handle.
- Every reader (`ThunderIssuerURL`, `ThunderExternalTokenURL/JWKSURL`, `EnvThunderResolver`, `ListThunderInstances`) now reads the stored `thunder_url` directly, nothing recomputes an origin at read time anymore.
- Added `ssrf`-hardened dialing (re-resolve at connect time, re-validate every redirect) for the one candidate that can carry a caller-supplied URL, cluster-internal and local-dev fallback candidates are untouched since those are never attacker-influenced.
- Regenerated the `amctl` CLI client and the migration job's Helm values, both derive from the same OpenAPI spec and schema change.

## User stories
As a SaaS control plane provisioning an environment, I can register its env-Thunder address directly by URL, without needing to reverse-engineer AMS's on-prem handle/domain convention.

## Release note
N/A

## Documentation
N/A

## Training
N/A

## Certification
N/A

## Marketing
N/A

## Automation tests
 - Unit tests
   `environment_service_unit_test.go` covers both registration paths: handle generation/collision retry, url validation and SSRF rejection, the both-set-is-400 case, and idempotent reuse-vs-reject for each path. `naming_test.go` and `client_test.go` prove the SSRF-hardened client rejects a loopback target while the plain one still reaches it. `make test-unit`: all packages pass.
 - Integration tests
   `env_thunder_url_repository_test.go` proves against a real Postgres that concurrent first-inserts race correctly, a handle collision and a url collision are reported as distinct conflicts, and multiple handle-less SaaS rows never collide with each other. `make test-integration`: all tests pass.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? no

## Samples
N/A

## Migrations (if applicable)
Migration 043 adds `thunder_url` (not null) and drops the not-null constraint on `thunder_handle`. Verified locally against Postgres 16 with both a fresh database and an upgrade path from an existing handle-only row.

## Test environment
macOS (Darwin), Docker Compose for agent-manager-service and Postgres, Go 1.25, Postgres 16 (postgres:16-alpine).

## Learning
N/A

## Summary by CodeRabbit

* **New Features**
  * Environment Thunder registration supports either an on-premises handle or a full caller-supplied URL.
  * Responses include the resolved URL and an optional handle.
  * URLs are validated, normalized, and protected against unsafe destinations.
* **Bug Fixes**
  * Added clear errors for invalid URLs, duplicate registrations, and conflicting inputs.
  * Existing registrations are migrated to resolved URLs.
* **Documentation**
  * Updated API and deployment guidance.
* **Chores**
  * Renamed the base-domain configuration setting to `idpHostBaseDomain`.


deployments/helm-charts/wso2-agent-manager/templates/jobs/db-migration-job.yaml
```diff
@@ -88,6 +88,14 @@ spec:
                   {{- end }}
             - name: IS_LOCAL_DEV_ENV
               value: "true"
+            # Migration 043 backfills existing handle-only env-Thunder rows by
+            # using the same deployment-specific origin settings as amp-api.
+            # Keep these in the migration hook as well as the API ConfigMap or
+            # VM/custom-domain upgrades silently persist localhost URLs.
+            - name: THUNDER_HOST_BASE_DOMAIN
```
menakaj:
```suggestion
            - name: IDP_HOST_BASE_DOMAIN
```

deployments/helm-charts/wso2-agent-manager/tests/render.sh
```diff
@@ -159,6 +159,16 @@ assert_env "sslRootCert reaches the migration job" "$MIG_TMPL" DB_SSL_ROOT_CERT
   "${EXTERNAL[@]}" --set postgresql.external.sslMode=verify-full \
   --set postgresql.external.sslRootCert=system
 
+# Migration 043 derives a full URL for existing handle-only rows. The hook must
+# receive the same deployment origin settings as the API or a TLS/custom-domain
+# upgrade would backfill localhost URLs even though live writes are correct.
+assert_env "Thunder base domain reaches the migration job" \
+  "$MIG_TMPL" THUNDER_HOST_BASE_DOMAIN "amp.example.com" \
```
menakaj:
```suggestion
  "$MIG_TMPL" IDP_HOST_BASE_DOMAIN "amp.example.com" \
```

deployments/helm-charts/wso2-agent-manager/templates/jobs/db-migration-job.yaml
```diff
@@ -88,6 +88,14 @@ spec:
...
+              value: {{ .Values.agentManagerService.config.thunderHostBaseDomain | default "amp.localhost" | quote }}
```
menakaj:
```suggestion
              value: {{ .Values.agentManagerService.config.idpHostBaseDomain | default "amp.localhost" | quote }}
```

menakaj (commented):
Replace user-facing ThunderID with IDP

deployments/helm-charts/wso2-agent-manager/values.yaml
```diff
@@ -254,10 +254,10 @@ agentManagerService:
     oauthAuthorizationServers: ""
 
     # Domain suffix env-Thunder developer-facing hostnames are built from:
-    # "<handle>.<thunderHostBaseDomain>". Override on custom-domain /
+    # "<handle>.<idpHostBaseDomain>". Override on custom-domain /
     # sslip.io installs so the reported identity-provider endpoints match the
     # actual env-Thunder hosts.
-    thunderHostBaseDomain: "amp.localhost"
+    idpHostBaseDomain: "amp.localhost"
```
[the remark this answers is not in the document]
yashed: Didnt add backward compatibility because of pre release

agent-manager-service/clients/thundersvc/naming.go
```diff
@@ -283,9 +292,9 @@ func resolveThunderBaseURL(ctx context.Context, org, env, handle string, prober
 // dev only) Docker Desktop/Linux host-networking fallbacks. Callers that build an HTTP
 // client against the result must dial resolveToHost (when non-empty) instead of the
 // base URL's own host, while still sending the base URL's host as the Host header.
-// See ThunderExternalTokenURL for handle.
-func ResolveThunderBaseURL(ctx context.Context, org, env, handle string) (baseURL, resolveToHost string, ok bool) {
-	c, ok := resolveThunderBaseURL(ctx, org, env, handle, defaultThunderURLProber)
+// thunderURL is the stored EnvThunderURL.ThunderURL origin.
+func ResolveThunderBaseURL(ctx context.Context, org, env, thunderURL string) (baseURL, resolveToHost string, ok bool) {
+	c, ok := resolveThunderBaseURL(ctx, org, env, thunderURL, defaultThunderURLProber)
```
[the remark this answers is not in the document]
yashed: fixed

deployments/helm-charts/wso2-agent-manager/values.schema.json
```diff
@@ -844,10 +844,10 @@
               "title": "oauthAuthorizationServers",
               "type": "string"
             },
-            "thunderHostBaseDomain": {
+            "idpHostBaseDomain": {
```
jhivandb, line 847: **The docs still set the old `thunderHostBaseDomain` key, which this schema now rejects.** _(high)_

This PR renames the values key to `idpHostBaseDomain` and touches no documentation files. Both `agentManagerService.config` and `console.config` are `"additionalProperties": false`, so the documented `helm upgrade` for a custom-domain/TLS install no longer silently ignores the stale key — it aborts with `Additional property thunderHostBaseDomain is not allowed`.

Stale call sites:
- `documentation/docs/guides/on-your-environment.mdx` lines 1410, 1412, 1440, 1442 — `--set agentManagerService.config.thunderHostBaseDomain=…` / `--set console.config.thunderHostBaseDomain=…`
- `documentation/docs/reference/helm-charts/wso2-agent-manager.mdx` lines 109 and 246 — the values table rows (and their descriptions, which still say `"<handle>.<thunderHostBaseDomain>"`)

The `THUNDER_HOST_BASE_DOMAIN` *environment variable* references elsewhere in those guides (lines 1481, 1530) are still correct — the scripts kept that name — so only the Helm `--set` keys and the reference table need changing.

agent-manager-service/services/environment_service.go
```diff
@@ -882,7 +1009,7 @@ func (s *environmentService) SetThunderSystemClientSecret(ctx context.Context, o
 	if ouID == "" {
 		return fmt.Errorf("%w: ouID is required", utils.ErrInvalidInput)
 	}
-	if _, err := s.SetThunderURL(ctx, ouID, envName, ""); err != nil {
+	if _, err := s.SetThunderURL(ctx, ouID, envName, "", ""); err != nil {
```
jhivandb: **A control plane that stores the system-client secret before registering the URL permanently stamps a bogus origin.** _(high)_

`SetThunderURL(ctx, ouID, envName, "", "")` takes the handle path for an unregistered environment: it auto-generates a handle and computes the origin from `IDP_HOST_BASE_DOMAIN` (default `amp.localhost`). That was sound when the handle was the only path, but the premise of this PR is that in the SaaS/control-plane flow there is no meaningful global base domain.

Nothing enforces call ordering — `add-environment-thunder.sh` happens to call `register_thunder_url` first, but that is exactly the "external, unenforced ordering" this method's own doc comment says it exists to fix. If any caller reaches `SetThunderSystemClientSecret` before `PUT .../thunder-url {"url": …}`, the environment is stamped with something like `http://a1b2c3d4e5.amp.localhost:8080`, and the real registration then fails forever with `ErrThunderURLTaken` → 409 until someone calls `DELETE`. Because `SetThunderURL` is explicitly "never changes an already-registered record", there is no recovery short of deletion.

The doc comment's claim that the invariant "hold[s] by construction for every caller, present and future" is no longer true for the URL path. Worth either gating the auto-provision on the deployment mode, or storing the credential without provisioning a URL when none exists yet.

agent-manager-service/services/environment_service.go
```diff
@@ -699,125 +708,243 @@ func generateThunderHandle() (string, error) {
...
+
+// validateThunderURL checks rawURL's shape for the SaaS/control-plane
+// registration path and returns it normalized to a bare "scheme://host[:port]"
+// origin (stripping any trailing "/", since callers append their own paths).
+// Must be an absolute http(s) origin with no userinfo/path/query/fragment.
+// Also runs ssrf.ValidateURL: unlike a handle (always AMS's own label under
+// its own trusted domain), this exact value gets dialed later by
+// EnvThunderResolver — genuinely attacker-influenced input, which a handle
+// never was.
+func validateThunderURL(ctx context.Context, rawURL string) (string, error) {
```
jhivandb, line 723: **The URL path skips the reserved-handle protection the handle path enforces.** _(medium)_

`validateThunderURL` checks scheme, host, userinfo, path, query, fragment and `ssrf.ValidateURL`, but never applies `reservedThunderHandles` (line 632). An org admin holding `OrgManageServiceAccount` can therefore register `https://console.<idpBaseDomain>`, `https://api.<idpBaseDomain>`, `https://thunder.<idpBaseDomain>`, `https://gateway.<idpBaseDomain>` and so on as their environment's Thunder origin — precisely the hijack/confusion case the handle path blocks by name.

The consequence is worse than for a handle: AMS then dials that platform host with the environment's system-client credentials on every admin/token call, and advertises it as the environment's issuer/token/JWKS on the console Identity page.

The check is not a straight reuse — the reserved set is single DNS labels, so it needs to compare against the first label of the host when the host sits under the configured base domain, rather than against the whole host.

agent-manager-service/clients/thundersvc/naming.go
```diff
@@ -202,16 +202,17 @@ type thunderURLCandidate struct {
...
+// provisioned" before ever reaching here).
+func thunderBaseURLCandidates(org, env, thunderURL string) []thunderURLCandidate {
+	if thunderURL == "" {
+		panic("thunder url must not be empty — callers must check for \"not provisioned\" before building candidates")
+	}
+	externalBaseURL := thunderURL
 	candidates := []thunderURLCandidate{
 		{baseURL: ThunderInternalURL(org, env)},
-		{baseURL: externalBaseURL},
+		{baseURL: externalBaseURL, external: true},
```
jhivandb: **`external: true` is unconditional, so legitimately-private on-prem origins get rejected by the SSRF guard.** _(medium)_

The `thunderURLCandidate` doc comment justifies the flag for caller-supplied SaaS values — genuinely attacker-influenced input. But this candidate is built from the stored `thunder_url` regardless of which path produced it, and on the handle path that value is AMS-computed under its own trusted base domain. The flag cannot tell the two apart.

`ssrf.NewClient` rejects any host resolving to a private/RFC1918/loopback address, and handle-derived origins can legitimately be private:
- a VM install using sslip.io on a LAN IP — `amp.10.0.0.5.sslip.io`, exactly the shape `deployments/vm/install-vm.sh` produces from the host IP
- an internal corporate DNS name for an on-prem deployment

On any deployment where the cluster-internal candidate also fails (AMS and env-Thunder in separate clusters), env-Thunder resolution goes from working to permanently `ErrThunderUnreachable`.

Since the row already records which path created it (`thunder_handle` non-null vs. null), carrying that down into the candidate list would classify correctly instead of inferring `external` from the candidate's position.

_Distinct from the resolved thread on line 297 — that one was about propagating `external` far enough; this is about it being applied too broadly._

agent-manager-service/clients/thundersvc/naming.go
```diff
@@ -240,7 +245,11 @@ func probeThunderURL(ctx context.Context, url, resolveToHost string) bool {
 		req.Host = req.URL.Host
 		req.URL.Host = resolveToHost
 	}
-	resp, err := http.DefaultClient.Do(req)
+	httpClient := http.DefaultClient
+	if external {
+		httpClient = ssrf.NewClient(probeTimeout)
```
jhivandb: **A fresh SSRF client per probe leaks a connection pool each call.** _(low)_

`ssrf.NewClient` does `http.DefaultTransport.(*http.Transport).Clone()` (`utils/ssrf/ssrf.go:169`), so every probe builds a brand-new `http.Transport` with its own connection pool, and nothing ever calls `CloseIdleConnections` on it. `ThunderProbe` runs once per environment on every `ListThunderInstances` — i.e. every load of the console Identity page — so idle connections and their reader goroutines accumulate for the transport's 90s idle timeout under any polling load.

A package-level `var ssrfProbeClient = ssrf.NewClient(probeTimeout)` is enough; the client is stateless and safe for concurrent use, and the per-request timeout already comes from `reqCtx`.

agent-manager-service/db_migrations/043_add_thunder_url_to_env_thunder_urls.go
```diff
@@ -0,0 +1,83 @@
...
+			}
+			if err := tx.Table("env_thunder_urls").
+				Select("id, thunder_handle").
+				Where("thunder_url IS NULL AND thunder_handle IS NOT NULL").
+				Find(&rows).Error; err != nil {
+				return fmt.Errorf("read pre-existing env_thunder_urls rows for backfill: %w", err)
+			}
+
+			for _, row := range rows {
+				thunderURL := thundersvc.ThunderOriginFromHandle(row.ThunderHandle)
```
jhivandb, line 74: **An empty-string handle panics the migration and wedges the upgrade at version 43.** _(low)_

The backfill filters `thunder_handle IS NOT NULL` but not `<> ''`, and `ThunderOriginFromHandle` panics outright on an empty handle (`naming.go:135`). Before this migration the column was `NOT NULL` with no `CHECK (thunder_handle <> '')`, so an empty string was a storable value — one such row panics the migration job and blocks the entire upgrade, with the transaction rolled back and nothing to retry into.

Adding `AND thunder_handle <> ''` to the `Where`, or skipping such rows explicitly, removes the sharp edge. Note the subsequent `SET NOT NULL` on `thunder_url` means a skipped row would then fail the ALTER, so an explicit error mentioning the offending row id is probably better than a silent skip.

jhivandb (commented):
## AI code review

Automated review of `main...HEAD` (`15dddea7`), verified against the current PR head. Six findings are inline below. Two notes that could not be anchored to a diff line:

**`agent-manager-service/services/environment_service.go:968` — low.** `IsThunderHandleAvailable` still consults only `ExistsByHandle`. Now that `uq_env_thunder_urls_url` is the primary invariant, the console's Create Environment drawer can report a handle as available while the origin it derives to (`https://<handle>.<idpHostBaseDomain>`) is already owned by a URL-path row. The pre-flight check then promises a registration that fails with 409 `ErrThunderURLTaken`.

**Already covered:** the host-case normalization at `environment_service.go:748` is the existing CodeRabbit thread; I have added the default-port half of it as a reply there rather than opening a duplicate.

**Checked and clean:** `ListThunderInstances` still guards the now-panicking `ThunderIssuerURL("")` behind `reachable[i]`, and `rec.URL == ""` short-circuits before `thunderBaseURLCandidates` can panic. The resolver's `resolveToHost == "" && baseURL == thunderURL` heuristic correctly selects only the plain-external candidate. Constraint names in `env_thunder_url_repository.go` match migrations 042/043 and no `AutoMigrate` collides with them. `add-environment-thunder.sh` / `ams-auth.sh` still parse `"handle"`, which the on-prem path always emits, and `add-environment.sh` does not read the renamed variable back from the ConfigMap.

agent-manager-service/services/environment_service.go
```diff
@@ -699,125 +708,243 @@ func generateThunderHandle() (string, error) {
...
+	if parsed.User != nil {
+		return "", fmt.Errorf("%w: must not contain userinfo", utils.ErrInvalidThunderURL)
+	}
+	if parsed.Path != "" && parsed.Path != "/" {
+		return "", fmt.Errorf("%w: must be a bare origin with no path", utils.ErrInvalidThunderURL)
+	}
+	if parsed.RawQuery != "" || parsed.Fragment != "" {
+		return "", fmt.Errorf("%w: must not contain a query or fragment", utils.ErrInvalidThunderURL)
+	}
+	origin := fmt.Sprintf("%s://%s", parsed.Scheme, parsed.Host)
```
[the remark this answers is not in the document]
jhivandb: **AI review addendum — the default port needs stripping too, not just the case.**

Beyond `Example.com` vs `example.com`, `parsed.Host` also keeps an explicitly-written default port. `https://idp.example.com` and `https://idp.example.com:443` (and `http://…` vs `http://…:80`) are the same origin but normalize to different stored strings, so `uq_env_thunder_urls_url` does not fire and two different orgs can each register the same real origin. That breaks the "globally unique across ALL orgs/environments" invariant the OpenAPI description and migration043's comment both state, and leaves two environments advertising an identical issuer.

Combined with the case fix:

```go
host := strings.ToLower(parsed.Host)
if (parsed.Scheme == "https" && strings.HasSuffix(host, ":443")) ||
    (parsed.Scheme == "http" && strings.HasSuffix(host, ":80")) {
    host = parsed.Hostname()
}
origin := fmt.Sprintf("%s://%s", parsed.Scheme, host)
```

(`parsed.Hostname()` also drops the IPv6 brackets, so re-add them if a literal-IPv6 origin is ever in scope.)

deployments/helm-charts/wso2-agent-manager/values.schema.json
yashed: updated the 6 stale key references in both docs files.

agent-manager-service/services/environment_service.go
yashed: removed the auto-provisioning call entirely. The resolver already handles "credential without URL" as a valid not-provisioned state, so nothing else needed to change.

agent-manager-service/services/environment_service.go
yashed: added the same reserved-word check, scoped to this deployment's own base domain so it won't false-positive on a real SaaS domain.

agent-manager-service/clients/thundersvc/naming.go
yashed: the flag was based on candidate position, not which path actually produced the URL. Fixed by threading the real handle/no-handle signal through the resolver chain.

agent-manager-service/clients/thundersvc/naming.go
yashed: moved to a shared package-level client

agent-manager-service/db_migrations/043_add_thunder_url_to_env_thunder_urls.go
yashed: added an explicit check that fails with the row ID instead of panicking. Reproduced it against a real DB to confirm.

agent-manager-service/services/environment_service.go
```diff
@@ -699,125 +710,255 @@ func generateThunderHandle() (string, error) {
...
+	// A url under THIS deployment's own base domain is still checked against
+	// reservedThunderHandles, the same protection the handle path enforces:
+	// without it, a caller could register e.g. "https://console.<baseDomain>"
+	// and AMS would dial the platform's own console with the environment's
+	// system-client credentials, and advertise it as the environment's
+	// issuer/JWKS. A url under a DIFFERENT domain (the normal SaaS case) never
+	// matches this deployment's base domain, so this never fires for it.
+	if label, rest, ok := strings.Cut(parsed.Hostname(), "."); ok &&
+		strings.EqualFold(rest, config.GetConfig().ThunderHostBaseDomain) &&
+		reservedThunderHandles[strings.ToLower(label)] {
```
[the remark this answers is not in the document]
yashed: Fixed

jhivandb (approved):

PR #1780, opened 2026-08-31
Move agent configuration from the Workload to per-environment release-bindings

### LINKED ISSUE #1690: [Bug]: Config set at agent creation stays component-wide and cannot be removed.

Area/Agent Lifecycle (Agent create, configure, kinds, build, test, deploy, keys/tokens)

Aspect/Agent Configuration (Agent configuration, kinds, build options)

### Description

The configurations provided during the create flow are appended to the component workload. Configurations set at the workload level are inherited by every environment. In each environment, we can override the configured values, but we cannot unset or remove them.

This would be an issue for Ballerina agents because the Ballerina code break if unused configurations are present.

### Steps to Reproduce

1. Create a Ballerina agent with an environment variable that is not used by the application.
2. Wait for the deployment to complete.
3. The deployment will get stuck in an error state, with an error simialr to below.
```
error: [BAL_CONFIG_VAR_OAIAPIKEY] unused environment variable
```

### Severity Level of the Issue

Severity/Major (Important functionality is broken. Should be prioritized. Doesn't need immediate attention)

### Environment Details (with versions)

_No response_

### LINKED ISSUE #1771: [Bug]: Deployment fails with "Failed to collect SecretReferences" after deleting a secret runtime variable

Area/Agent Lifecycle (Agent create, configure, kinds, build, test, deploy, keys/tokens)

Aspect/Agent Configuration (Agent configuration, kinds, build options)

### Description

When a runtime variable configured as a secret is deleted, the backend releasebinding controller throws a Failed to collect SecretReferences error. This results in a Reconciler error and causes the deployment process to halt and fail completely.

**Error Log Snippet:**

```json
2026-08-31T05:58:38Z    ERROR    Failed to collect SecretReferences    {"controller": "releasebinding", "controllerGroup": "openchoreo.dev", "controllerKind": "ReleaseBinding", "ReleaseBinding": {"name":"citizen-inqui...
2026-08-31T05:58:38Z    ERROR    Reconciler error    {"controller": "releasebinding", "controllerGroup": "openchoreo.dev", "controllerKind": "ReleaseBinding", "ReleaseBinding": {"name":"citizen-inquiry-agent-he-defaul...

```

**Steps to Reproduce***

1. Configure an agent/component with a runtime variable that is marked as a secret.
2. Delete the secret runtime variable from the configuration.
3. Observe that the deployment errors out, and the backend logs surface the `Failed to collect SecretReferences` reconciliation error.

### Steps to Reproduce

mentioned

### Severity Level of the Issue

Severity/Blocker (Core functionality is broken and consumer is blocked. Need immediate attention)

### Environment Details (with versions)

latest make setup

hanzjk:
The Workload now carries just the image and the endpoint contract. Configuration moves to `spec.workloadOverrides` on the per-environment ReleaseBinding.

**Source-built agents.** `amp-generate-workload` step in the build workflow gains two steps: cut the ComponentRelease, then create or update the first environment's ReleaseBinding with the configuration as `workloadOverrides`. A rebuild advances only `releaseName` — once the binding exists it owns that environment's configuration, and a rebuild must not roll it back to whatever the component-level parameters happened to hold.

**Kind-sourced agents.** These run no build workflow, so the backend does the same work in `CreateKindAgentReleaseAndBinding`.

**Authorization.** Cutting a release and writing a binding needs actions OpenChoreo's built-in `workload-publisher` role does not grant, so a supplementary `ClusterAuthzRole` covers `componentrelease: create`,
`releasebinding` view/create/update, and the `project` / `deploymentpipeline` reads used to resolve the first environment. The binding matches on the `client_id` claim, as every other machine-client binding in the system does.

Resolves https://github.com/wso2/agent-manager/issues/1771
Resolves https://github.com/wso2/agent-manager/issues/1690 

## Goals
An agent's env vars and file mounts were written onto its Workload CR. A value placed there is inherited by every environment, and while a ReleaseBinding's `workloadOverrides` can *replace* such a value per environment, it cannot
*remove* it — so the inheritance is impossible to opt out of. An environment that should not carry a key has no way to say so.

Make configuration environment-scoped: the Workload carries only what is genuinely environment-independent, and each environment's configuration lives on its own ReleaseBinding. Apply this to both agent sources — source-built and
kind-sourced — so the two paths resolve configuration the same way.

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary by CodeRabbit

- **New Features**
  - Environment-specific variables and file mounts are now applied through release bindings, preserving configuration independently for each deployment environment.
  - Agent deployments automatically create and bind releases to the initial environment.
  - Rebuilding workloads updates deployment bindings while retaining environment-specific configuration.
  - Secret references are recognized from both shared workloads and environment-specific configuration.
  - Added permissions to support automated release and binding management.

- **Bug Fixes**
  - Prevented environment configuration from being stored on shared workloads, avoiding unintended cross-environment changes.

- **UI Improvements**
  - Simplified the agent promotion configuration panel by removing outdated guidance.

### DIFF
--- a/agent-manager-service/clients/openchoreosvc/client/components.go
+++ b/agent-manager-service/clients/openchoreosvc/client/components.go
@@ -123,7 +123,14 @@ func buildInternalAgentFromKindComponentRequestBody(namespaceName, projectName s
 		return gen.CreateComponentJSONRequestBody{}, fmt.Errorf("failed to convert parameters to map: %w", err)
 	}
 
-	autoDeploy := true
+	// Kind-sourced agents have no build workflow, so nothing runs amp-generate-workload
+	// to cut the ComponentRelease and bind it to the first environment. The backend does
+	// that itself instead (CreateKindAgentReleaseAndBinding), for the same reason the
+	// workflow does it for source-built agents: configuration must land on the
+	// per-environment ReleaseBinding rather than the shared Workload, and autoDeploy's
+	// controller-created binding carries no overrides. Both agent kinds therefore set
+	// this to false and own the release/binding themselves.
+	autoDeploy := false
 	return gen.CreateComponentJSONRequestBody{
 		Metadata: gen.ObjectMeta{
 			Name:        req.Name,
@@ -249,7 +256,12 @@ func buildInternalAgentFromSourceComponentRequestBody(namespaceName, projectName
 		return gen.CreateComponentJSONRequestBody{}, fmt.Errorf("error building workflow parameters: %w", err)
 	}
 
-	autoDeploy := true
+	// The build workflow's generate-workload step cuts the ComponentRelease and binds
+	// it to the first environment itself, which is what lets each environment's
+	// configuration live on its own ReleaseBinding instead of on the shared Workload.
+	// autoDeploy would have OpenChoreo's Component controller create and own that same
+	// binding, so the two would fight over spec.releaseName.
+	autoDeploy := false
 	return gen.CreateComponentJSONRequestBody{
 		Metadata: gen.ObjectMeta{
 			Name:        req.Name,
--- a/agent-manager-service/clients/openchoreosvc/client/components.go
+++ b/agent-manager-service/clients/openchoreosvc/client/components.go
@@ -125,7 +125,7 @@ func buildInternalAgentFromKindComponentRequestBody(namespaceName, projectName s
 
 	// Kind-sourced agents have no build workflow, so nothing runs amp-generate-workload
 	// to cut the ComponentRelease and bind it to the first environment. The backend does
-	// that itself instead (CreateKindAgentReleaseAndBinding), for the same reason the
+	// that itself instead (EnsureReleaseAndBinding), for the same reason the
 	// workflow does it for source-built agents: configuration must land on the
 	// per-environment ReleaseBinding rather than the shared Workload, and autoDeploy's
 	// controller-created binding carries no overrides. Both agent kinds therefore set
@@ -173,6 +173,10 @@ func buildExternalAgentComponentRequestBody(namespaceName, projectName string, r
 		return gen.CreateComponentJSONRequestBody{}, err
 	}
 
+	// An external agent has no Workload and never deploys, so autoDeploy has nothing to act on.
+	// Stated rather than left to the field's default, so all three component builders answer the
+	// question in the same place.
+	autoDeploy := false
 	return gen.CreateComponentJSONRequestBody{
 		Metadata: gen.ObjectMeta{
 			Name:        req.Name,
@@ -193,6 +197,7 @@ func buildExternalAgentComponentRequestBody(namespaceName, projectName string, r
 			}{
 				ProjectName: projectName,
 			},
+			AutoDeploy: &autoDeploy,
 		},
 	}, nil
 }
@@ -257,10 +262,11 @@ func buildInternalAgentFromSourceComponentRequestBody(namespaceName, projectName
 	}
 
 	// The build workflow's generate-workload step cuts the ComponentRelease and binds
-	// it to the first environment itself, which is what lets each environment's
-	// configuration live on its own ReleaseBinding instead of on the shared Workload.
-	// autoDeploy would have OpenChoreo's Component controller create and own that same
-	// binding, so the two would fight over spec.releaseName.
+	// it to the first environment itself, and every deploy afterwards does the same via
+	// EnsureReleaseAndBinding. That is what lets each environment's configuration live on
+	// its own ReleaseBinding instead of on the shared Workload. autoDeploy would have
+	// OpenChoreo's Component controller create and own that same binding, so the two would
+	// fight over spec.releaseName.
 	autoDeploy := false
 	return gen.CreateComponentJSONRequestBody{
 		Metadata: gen.ObjectMeta{
--- a/agent-manager-service/clients/openchoreosvc/client/deployments.go
+++ b/agent-manager-service/clients/openchoreosvc/client/deployments.go
@@ -106,9 +106,13 @@ func (c *openChoreoClient) CreateInternalAgentFromKindWorkload(ctx context.Conte
 		endpointMap[name] = workloadEp
 	}
 
-	envVars := toGenEnvVars(req.Env)
-	fileVars := toGenFileVars(req.Files)
-
+	// Env vars and file mounts are deliberately NOT set here. A value on the Workload is
+	// inherited by every environment, and a ReleaseBinding's workloadOverrides can replace
+	// such a value per environment but cannot remove it — making the inheritance impossible
+	// to opt out of. Configuration therefore lives only on the per-environment binding,
+	// written by CreateKindAgentReleaseAndBinding; the Workload carries just the image and
+	// the environment-independent endpoint contract. This mirrors amp-generate-workload,
+	// which splits the same two documents for source-built agents.
 	workload := gen.CreateWorkloadJSONRequestBody{
 		Metadata: gen.ObjectMeta{
 			Name:      workloadName,
@@ -117,8 +121,6 @@ func (c *openChoreoClient) CreateInternalAgentFromKindWorkload(ctx context.Conte
 		Spec: &gen.WorkloadSpec{
 			Container: &gen.WorkloadContainer{
 				Image: req.ImageID,
-				Env:   &envVars,
-				Files: &fileVars,
 			},
 			Owner: &struct {
 				ComponentName string `json:"componentName"`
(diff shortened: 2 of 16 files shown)


deployments/helm-charts/wso2-amp-platform-resources-extension/templates/cluster-workflow-templates/generate-workload.yaml
```diff
@@ -297,6 +316,192 @@ spec:
...
+                printf '%s' "${BINDING_GET_BODY}" > /mnt/vol/releasebinding-get.json
+                jq -c --arg releaseName "$RELEASE_NAME" '.spec.releaseName = $releaseName' \
+                  /mnt/vol/releasebinding-get.json > /mnt/vol/releasebinding.json
+
+                BINDING_RESPONSE=$(curl -s -w "\n%{http_code}" \
+                  -X PUT "${API_URL}/api/v1/namespaces/${NAMESPACE}/releasebindings/${BINDING_NAME}" \
+                  -H "Host: ${API_HOST}" \
+                  -H "Content-Type: application/json" \
+                  -H "Authorization: Bearer ${ACCESS_TOKEN}" \
+                  -d @/mnt/vol/releasebinding.json)
```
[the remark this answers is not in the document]
hanzjk: both proposed remedies are unavailable in OC API

agent-manager-service/clients/openchoreosvc/client/components.go
```diff
@@ -249,7 +256,12 @@ func buildInternalAgentFromSourceComponentRequestBody(namespaceName, projectName
 		return gen.CreateComponentJSONRequestBody{}, fmt.Errorf("error building workflow parameters: %w", err)
 	}
 
-	autoDeploy := true
+	// The build workflow's generate-workload step cuts the ComponentRelease and binds
+	// it to the first environment itself, which is what lets each environment's
+	// configuration live on its own ReleaseBinding instead of on the shared Workload.
+	// autoDeploy would have OpenChoreo's Component controller create and own that same
+	// binding, so the two would fight over spec.releaseName.
+	autoDeploy := false
```
jhivandb, line 270: **Turning off `autoDeploy` removes the only mechanism that cut a new release on an image change, so `DeployAgent` no longer changes what runs.**

`DeployAgent` → `ocClient.Deploy()` (`deployments.go:148-204`) only writes `workload.Spec.Container.Image`; it never cuts a `ComponentRelease` or advances `spec.releaseName`. AutoDeploy was what did that. After this change the backend's only `GenerateRelease` call is the new kind-creation path, and the build workflow's is the only other one.

Failure scenario: a user picks an older build in `BuildSelectorDrawer` and clicks **Configure & Deploy**. The Workload image is updated, but the binding stays pinned to the release cut by the last build workflow — so the pod keeps serving the previous image while the API returns 202 and an `agent.deploy` audit record is written. Same shape at `components.go:133` for a kind agent redeployed onto a new kind version.

Either `Deploy` needs to cut a release and repoint the binding the way `CreateKindAgentReleaseAndBinding` does, or the deploy path needs to be routed through the workflow.

agent-manager-service/clients/openchoreosvc/client/deployments.go
```diff
@@ -106,9 +106,13 @@ func (c *openChoreoClient) CreateInternalAgentFromKindWorkload(ctx context.Conte
 		endpointMap[name] = workloadEp
 	}
 
-	envVars := toGenEnvVars(req.Env)
-	fileVars := toGenFileVars(req.Files)
-
+	// Env vars and file mounts are deliberately NOT set here. A value on the Workload is
```
jhivandb, line 109: **Agent deletion no longer cleans up secrets.**

`DeleteAgent` discovers secret references only via `GetWorkloadSecretRefNames` (`agent_manager.go:2609`), which scans `workload.Spec.Container.Env[].valueFrom.secretKeyRef` (`secret_references.go:282-288`). Now that neither this function nor `generate-workload.yaml` writes env onto the Workload, that scan always returns `[]` and `cleanupSecretReference` never runs.

Failure scenario: create an agent with a sensitive env var (`saveSecretsAndCreateReference` stores it in KV and creates a SecretReference CR), then delete the agent — the KV secret and the CR are orphaned permanently. Nothing else in `DeleteAgent` cleans them up, and the integration test mocks `GetWorkloadSecretRefNames`, so no test catches it.

The discovery needs to move to the ReleaseBindings' `workloadOverrides`, across every environment the agent is bound to.

agent-manager-service/services/agent_manager.go
```diff
@@ -1693,6 +1694,29 @@ func (s *agentManagerService) createComponentAgent(ctx context.Context, ouID, pr
...
 			s.logger.Info("Created internal-agent-from-kind workload", "agentName", req.Name)
+
+			// Cut the release and bind it to the first environment, carrying the resolved
+			// configuration as the binding's workloadOverrides. This is what
+			// amp-generate-workload does for source-built agents; kind components are
+			// created with autoDeploy off so nothing else deploys them.
+			if err := s.ocClient.CreateKindAgentReleaseAndBinding(
+				ctx, ouID, projectName, req.Name, firstEnv, kindEnvVars, kindFileVars,
+			); err != nil {
+				s.logger.Error("Failed to create release binding for kind-sourced agent",
```
jhivandb, line 1701: **This rollback path is narrower than `rollbackAgentCreate`, so it leaks DB rows.**

It calls `cleanupSecretsOnRollback` + `DeleteComponent`, but not `rollbackAgentCreate` — which exists a few lines above and already handles the `agentAPIArtifact` row created earlier in this same function (as well as the LLM/MCP config records). When the release-binding step fails, the component is torn down but those rows are left orphaned.

Calling `rollbackAgentCreate` here instead would keep the two failure paths in the same function consistent.

agent-manager-service/clients/openchoreosvc/client/deployments.go
```diff
@@ -472,6 +474,131 @@ func (c *openChoreoClient) ReplaceReleaseBindingWorkloadOverrides(ctx context.Co
...
+	// Step 2: bind the release to the environment. A binding can already exist when an
+	// agent is recreated over a previous one, so update rather than fail on that path;
+	// retryReleaseBindingUpdate re-reads per attempt, as the backend races other writers
+	// of this same binding (MCP env vars, runtime class).
+	bindingName := componentName + "-" + environment
+	existing, err := c.findReleaseBindingForEnv(ctx, namespaceName, componentName, environment)
+	if err != nil {
+		return fmt.Errorf("failed to look up release binding for component %q in environment %q: %w",
+			componentName, environment, err)
+	}
```
jhivandb, line 551: **The existing-binding branch never resets `Spec.State`, while the create path below explicitly sets `Active`.**

If the adopted binding (agent recreated over a previous one, or the 409 branch) is in `Suspended`, the new agent gets its release pinned and its overrides written but stays suspended — no pods come up, and nothing in the response signals it. The create path three lines down sets `ReleaseBindingSpecStateActive` unconditionally, so the two diverge silently depending on whether a binding already existed.

Setting `rb.Spec.State = Active` in the mutator would make adoption match creation.

deployments/helm-charts/wso2-amp-platform-resources-extension/templates/cluster-workflow-templates/generate-workload.yaml
```diff
@@ -297,6 +316,192 @@ spec:
...
+                  --arg environment "$FIRST_ENV" \
+                  --arg releaseName "$RELEASE_NAME" \
+                  --slurpfile overrides "$OUT_OVERRIDES_JSON" '
+                    {
+                      metadata: { name: $name, namespace: $namespace },
+                      spec: (
+                        {
+                          owner: { projectName: $projectName, componentName: $componentName },
+                          environment: $environment,
+                          releaseName: $releaseName
```
jhivandb, line 444: **This create omits `spec.state`, unlike every Go create path.**

`CreateKindAgentReleaseAndBinding` and the other Go binding-create paths all set `state: Active` explicitly; this one relies on the CRD's default instead. That default isn't visible from this repo, so the two paths agree only by assumption — and if it ever changes (or is unset), source-built agents come up suspended while kind agents don't.

Adding `state: "Active"` to the `spec` object makes the workflow self-describing and matches the Go behaviour.

deployments/helm-charts/wso2-amp-platform-resources-extension/templates/cluster-workflow-templates/generate-workload.yaml
```diff
@@ -297,6 +316,192 @@ spec:
...
+                  /mnt/vol/releasebinding-get.json > /mnt/vol/releasebinding.json
+
+                BINDING_RESPONSE=$(curl -s -w "\n%{http_code}" \
+                  -X PUT "${API_URL}/api/v1/namespaces/${NAMESPACE}/releasebindings/${BINDING_NAME}" \
+                  -H "Host: ${API_HOST}" \
+                  -H "Content-Type: application/json" \
+                  -H "Authorization: Bearer ${ACCESS_TOKEN}" \
+                  -d @/mnt/vol/releasebinding.json)
+              else
+                echo "Failed to look up ReleaseBinding (HTTP ${BINDING_GET_HTTP_CODE}): ${BINDING_GET_BODY}"
```
jhivandb, line 485: **A transient GET failure fails the build without using the retry budget.**

The retry loop only retries *write* failures (409/5xx, further down). A transient 500/503 on `api_get .../releasebindings/${BINDING_NAME}` falls into this `else` and `exit 1`s immediately — failing the whole build even though `BINDING_ATTEMPT` is still 1 of 3.

Treating a 5xx on the read like a 5xx on the write (increment, sleep, `continue`) and reserving `exit 1` for genuinely unexpected codes would make the loop do what its own retry budget implies.

jhivandb (commented):
## 🤖 AI code review

Automated review of the two commits on this branch (`6f2b9a543..8a0f0d07a`). `go build ./...` passes and the new Helm template renders; the findings below are behavioural, not build breaks.

**6 findings** — 2 that look like real regressions in deploy and delete, 4 smaller robustness/consistency issues. Inline comments follow.

Also worth a look, not filed as findings:
- `InternalAgentFromKindWorkloadRequest.Env` / `.Files` are now dead fields — nothing reads them, so a future caller populating them would get silent data loss rather than a compile error.
- Comments on `applyEnvScopedWorkloadConfig`, `EnsureReleaseBindingRuntimeClass` and `releaseBindingWaitAttempts` still describe AutoDeploy as the thing that creates bindings, which is no longer true.

deployments/helm-charts/wso2-amp-platform-resources-extension/templates/cluster-workflow-templates/generate-workload.yaml
hanzjk: Addressed via https://github.com/wso2/agent-manager/pull/1780/commits/9688ee71ccd051b6cbe8c25724bfbeeae3e5d8bb

deployments/helm-charts/wso2-amp-platform-resources-extension/templates/cluster-workflow-templates/generate-workload.yaml
hanzjk: Addressed via https://github.com/wso2/agent-manager/pull/1780/commits/9688ee71ccd051b6cbe8c25724bfbeeae3e5d8bb

agent-manager-service/clients/openchoreosvc/client/deployments.go
hanzjk: Addressed via https://github.com/wso2/agent-manager/pull/1780/commits/9688ee71ccd051b6cbe8c25724bfbeeae3e5d8bb

agent-manager-service/services/agent_manager.go
hanzjk: Addressed via https://github.com/wso2/agent-manager/pull/1780/commits/9688ee71ccd051b6cbe8c25724bfbeeae3e5d8bb

agent-manager-service/clients/openchoreosvc/client/deployments.go
hanzjk: Addressed via https://github.com/wso2/agent-manager/pull/1780/commits/9688ee71ccd051b6cbe8c25724bfbeeae3e5d8bb

agent-manager-service/clients/openchoreosvc/client/components.go
hanzjk: Addressed via https://github.com/wso2/agent-manager/pull/1780/commits/9688ee71ccd051b6cbe8c25724bfbeeae3e5d8bb

agent-manager-service/clients/openchoreosvc/client/secret_references.go
```diff
@@ -251,10 +251,53 @@ func (c *openChoreoClient) DeleteSecretReference(ctx context.Context, ouID, secr
...
 	namespaceName := c.NamespaceFor(ouID)
-	// List workloads filtered by component
+
+	// Collect unique secret reference names across both sources.
+	secretRefNames := make(map[string]struct{})
+
+	// Source 1: per-environment configuration on the release bindings.
+	bindingsResp, err := c.ocClient.ListReleaseBindingsWithResponse(ctx, namespaceName, &gen.ListReleaseBindingsParams{
+		Component: &componentName,
+		Limit:     &defaultListLimit,
```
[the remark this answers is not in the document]
hanzjk: For all oc apis, pagination is not implemented with cursors

jhivandb (approved):

